Issue #328, opened 2021-06-04
Implement running local cluster via CLI


pkosiec:
## Description

1. **CLI:** Implement command to quickly set up local Capact cluster. It might be split into two commands (see the "Possible options" section
1. **Website:** Use [https://github.com/asciinema/asciinema-player](https://github.com/asciinema/asciinema-player)  or [https://github.com/ines/termynal](https://github.com/ines/termynal) to display command for bootstrapping local Capact

Tasks:

- [x] Provision kind cluster - https://github.com/capactio/capact/pull/372
- [x] Install Capact
- [ ]  Use `capat env create kind` and `capact install` in our scripts

### Possible option

- Migrate our bash scripts to Go. Check before implementation if that is possible,  if not then fallback to previous idea (clone repo and use Bash scripts)
   * Clone subdirectory hack [https://stackoverflow.com/questions/600079/how-do-i-clone-a-subdirectory-only-of-a-git-repository](https://stackoverflow.com/questions/600079/how-do-i-clone-a-subdirectory-only-of-a-git-repository|smart-link)
 * Use charts from latest/stable helm repo
 * Provision command uses kind Go API
 * Git, kubectl, helm installed beforehand if not installed
    - Use SKIP_DEPS_INSTALLATION=true
    - For Git, we can check whether we can use go-git or GitHub CLI Golang API
    - Implement installation of kubectl, helm in our installation scripts
 * Add flag to specify release version (tag on git repo - by default latest release?)
 * Modify installation script to allow not building images
 * Update doc
     - development guide
     - local installation doc
 * Use it for local development
 * Use CLI on CI
    - PR: build cli from sources and  use it for integration tests
    - Branch: use CLI for integration tests

### Ideas

```bash
capact provision kind
capact install # by default latest semver git tag,
# alternatively
capact install --version @main/0.3.0/0.5.0
```

## Reason

Currently, there is no way to simply install Capact on local machine. You need to clone the repo, install required dependencies and run `make dev-cluster`. The steps should be improved and simplified.

This will help with getting users to get started quickly with the platform

 1. For development
 1. And experimentation / POC

Trojan295:
Right now, I see a few open points:
1. If we want to use it also for development / CI set up, then we have to remember that our current scripts also build Docker images and upload them to the kind cluster to the flow would need to be:
   1. Build Docker images
   1. Create kind cluster using `capact provision kind`
   1. Upload Docker images to kind cluster
   1. Install Capact using `capact install`
   
[...]

PR #372, opened 2021-06-25
Add capact env create kind command

lukaszo:
## Description

- Add `capact cluster create kind` command

## Testing
Build:

```bash
go build -ldflags "-s -w" -o bin/capact ./cmd/cli/main.go                                                                                                                           
```

Run 

```bash
./bin/capact env create kind
```

Deploy Capact

```bash
make dev-cluster-update
```

Delete deployment

```bash
capact env delete kind
```

Make sure that Capact and its components are running ok.

## Related issue(s)

#328


cmd/cli/cmd/deprovision/deprovision.go
```diff
@@ -0,0 +1,15 @@
+package deprovision
+
+import "github.com/spf13/cobra"
+
+func NewCmd() *cobra.Command {
+	root := &cobra.Command{
+		Use:   "deprovision",
+		Short: "This command consists of multiple subcommands to Deprovision Kubernetes cluster",
```
mszostok, line 8:
```suggestion
		Short: "This command consists of multiple subcommands to deprovision Kubernetes cluster",
```

cmd/cli/cmd/deprovision/kind.go
```diff
@@ -0,0 +1,22 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "kind",
+		Short: "Delete local kind cluster",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return deprovision.Kind(cmd.Context(), opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.Name, "name", "kind-dev-capact", "cluster name, overrides KIND_CLUSTER_NAME, config (default kind-dev-capact)")
```
mszostok, line 19: `default` message is already generated by cobra.
```suggestion
	cmd.Flags().StringVar(&opts.Name, "name", "kind-dev-capact", "cluster name, overrides KIND_CLUSTER_NAME, config")
```

mszostok: setting env `KIND_CLUSTER_NAME` doesn't work for me

mszostok (commented):
General things:
- I know that in task we have `capact provision kind` but WDYT about changing that to `capact cluster create kind` IMO this is better as we already have `capact hub` `capact action`, `capact config` etc. and it better as it groups commands. 
- IMO we should embed the default config that we store under `hack/cluster-config/kind/config.yaml`. We want to ensure that the only thing that you need to do is to download Capact CLI and later execute `capact provision kind`, `capact install`. Downloading another file and always running `capact provision kind --cluster-config ./tmp/file.yaml` is cumbersome.

lukaszo:
> General things:
> 
> * I know that in task we have `capact provision kind` but WDYT about changing that to `capact cluster create kind` IMO this is better as we already have `capact hub`, `capact config` etc. and it better as it groups commands.

Agree, then we can also have a `capact cluster delete` which sound nicer than `deprovision` :+1: 

> * IMO we should embed the default config that we store under `hack/cluster-config/kind/config.yaml`. We want to ensure that the only thing that you need to do is to download Capact CLI and later execute `capact provision kind`, `capact install`. Downloading another file and always running `capact provision kind --cluster-config ./tmp/file.yaml` is cumbersome.

Good idea.

cmd/cli/cmd/deprovision/kind.go
lukaszo: copy paste from Kind. Do we need to support env variables?

mszostok: I don't thinks so :) nice but not required

cmd/cli/cmd/cluster/cluster.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"capact.io/capact/cmd/cli/cmd/cluster/create"
+	deletecluster "capact.io/capact/cmd/cli/cmd/cluster/delete"
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
+	hub := &cobra.Command{
+		Use:   "cluster",
+		Short: "This command consists of multiple subcommands to interact with Kuberentes cluster.",
```
mszostok, line 12:
```suggestion
		Short: "This command consists of multiple subcommands to interact with a Kubernetes cluster",
```

cmd/cli/cmd/cluster/create/create.go
```diff
@@ -0,0 +1,15 @@
+package create
+
+import "github.com/spf13/cobra"
+
+func NewCmd() *cobra.Command {
+	root := &cobra.Command{
+		Use:   "create",
+		Short: "This command consists of multiple subcommands to create Kubernetes cluster",
```
mszostok, line 8:
```suggestion
		Short: "This command consists of multiple subcommands to create a Kubernetes cluster",
```

cmd/cli/cmd/cluster/delete/kind.go
```diff
@@ -0,0 +1,22 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "kind",
+		Short: "Delete local kind cluster",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return delete.Kind(cmd.Context(), opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.Name, "name", "kind-dev-capact", "cluster name, overrides KIND_CLUSTER_NAME, config (default kind-dev-capact)")
```
mszostok, line 19:
```suggestion
	cmd.Flags().StringVar(&opts.Name, "name", "kind-dev-capact",  "cluster name, overrides config")
```

cmd/cli/cmd/cluster/create/kind.go
```diff
@@ -0,0 +1,29 @@
...
+		Use:   "kind",
+		Short: "Provision local kind cluster",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return create.Kind(cmd.Context(), opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.Name, "name", "kind-dev-capact", "cluster name, overrides config")
+	cmd.Flags().StringVar(&opts.Config, "cluster-config", "", "path to a kind config file")
+	cmd.Flags().StringVar(&opts.ImageName, "image", "", "node docker image to use for booting the cluster")
```
mszostok, line 23:
```suggestion
	cmd.Flags().StringVar(&opts.ImageName, "image", "kindest/node:v1.19.1", "node docker image to use for booting the cluster")
```

Currently we use the `kindest/node:v1.19.1` and now cluster installs `v1.21.1`. This can be problematic, as some resources were deprecated and removed after 1.19, so u may run into an issue in next task


BTW not sure if this is a good place, probably we should have it close to our default config :)

mszostok (approved):
Please address my comments before merging, but besides that it LGTM 👍

platten:
> General things:
> 
> * I know that in task we have `capact provision kind` but WDYT about changing that to `capact cluster create kind` IMO this is better as we already have `capact hub`, `capact config` etc. and it better as it groups commands.
> * IMO we should embed the default config that we store under `hack/cluster-config/kind/config.yaml`. We want to ensure that the only thing that you need to do is to download Capact CLI and later execute `capact provision kind`, `capact install`. Downloading another file and always running `capact provision kind --cluster-config ./tmp/file.yaml` is cumbersome.

So my thoughts, Kubernetes is just one sample deployment target (the only one we support now). I think that our CLI should represent how we see capact being used in the future. So maybe: capact target create kind [--<configurationParam1>=<configValue1>...]. This way we are telling capact to bootstrap a local kubernetes/kind target (for running capact), and in the future we could use the same pattern for bootstrapping eks/gke/rke/etc capact target deployments in the future.

Thoughts?

lukaszo:
> So maybe: capact target create kind

I'm ok with `target`. @mszostok  @pkosiec  any thoughts?

pkosiec:
Target is a bit too generic and not as meaningful as `cluster` for me. Can we keep the `cluster` as this is also universal? Or, as an alternative, we could use `environment`. But honestly, I would vote for `cluster`.

mszostok:
Same as @pkosiec, personally, I think that `cluster` still works well for eks/gke/rke2 etc. installations.

platten:
Here's the thing: cluster means many machines. Besides, it could be kubernetes, but it could be a group. Capact eventually should be agnostic when it comes to kubernetes/a single static binary or say snap/serverless or any other type of deployment. That said, I would be ok with `target cluster kind` or `target k8s kind`.

pkosiec:
Fair enough - I would vote for `environment` then 👍

mszostok:
tbh `capact target cluster create kind` is too long IMO, so if we need to change it, the `environment` name is better with alias `env`, so `capact environment create kind`, `capact env create kind`.

mszostok (approved):
🚀

Issue #334, opened 2021-06-08
Empty error message from OCH GraphQL client


pkosiec:
## Description

Sometimes on our long-running cluster, we get an empty error message from the GraphQL client we use:

``` while deleting TypeInstance with ID f6ca4548-ef05-4a0f-87c0-ac294b9b4dec: while executing mutation to delete TypeInstance: All attempts fail:

# 1: graphql:

```

We need to investigate whether there is a bug in our code (OCH client) or in our dependency (`retry-go`, `machinebox/graphql` client, etc.).

- This may not be related just to local OCH client. Investigate all GraphQL clients (Public OCH, Local OCH, Engine) for potential causes.

## Expected behavior

We get the error message for the root cause of GraphQL request failure.

## Actual behavior

Example run: [https://github.com/capactio/capact/runs/2762064610?check_suite_focus=true](https://github.com/capactio/capact/runs/2762064610?check_suite_focus=true)

``` Action Action execution should pick proper Implementation and inject TypeInstance based on cluster policy /capact.io/capact/test/e2e/action_test.go:58 STEP: Preparing input Type Instances STEP: 1. Expecting Implementation A is picked based on test policy and requirements met... Jun 7 08:37:28.515: while deleting TypeInstance with ID f6ca4548-ef05-4a0f-87c0-ac294b9b4dec: while executing mutation to delete TypeInstance: All attempts fail:

# 1: graphql:

Jun 7 08:37:29.539: while deleting TypeInstance with ID a0ecd4c3-c55a-497a-8c68-05b5e6242cb8: while executing mutation to delete TypeInstance: All attempts fail:

# 1: graphql:

• Failure [305.075 seconds] Action /capact.io/capact/test/e2e/action_test.go:37 Action execution /capact.io/capact/test/e2e/action_test.go:57 should pick proper Implementation and inject TypeInstance based on cluster policy [It] /capact.io/capact/test/e2e/action_test.go:58 Timed out after 300.000s. Expected : FAILED to be an element of <[]interface *** | len:1, cap:1>: [["SUCCEEDED"]] /capact.io/capact/test/e2e/action_test.go:343 ```

**Comment by mszostok** [comment]
We need to bump prio for this task. It's problematic when it comes to debugging. We had such problem in our engine:
```
[...]

PR #378, opened 2021-06-29
Add name for each query and mutation in Hub Go client

mszostok:
## Description

After bumping graphql-js dependency, the e2e tests are failing. After debugging, I detected that for all unnamed queries/mutations, the GraphQL API returns `400` status code. Unfortunately, I cannot detect how our CI on PR didn't detect that.

- Add name for each query and mutation in Hub Go client

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -44,6 +44,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Setup environment
         run: |
           . ./hack/ci/setup-env.sh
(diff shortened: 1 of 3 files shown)


pkosiec (approved):

pkosiec:
This is OK change,  I was able to confirm the issue on our stage cluster, but the response in GraphQL Playground is terrible:

```graphql
query {
  interfaces {
    path
  }
}
```
results in:
```
{
  "data": {},
  "errors": [
    {} // empty error!
  ]
}
```

and

```graphql
query Interfaces {
  interfaces {
    path
  }
}
```

in:
```
{
  "data": {
    "interfaces": [
      {
        "path": "cap.core.interface.runner.generic.run"
      },
      {
// (...)
```

IMHO we should handle the issue #334 really soon.

Also, on this PR I noticed another, very similar fail on integration tests:

```
  Timed out after 300.001s.
  Error: Unexpected non-nil/non-zero extra argument at index 1:
  	<*errors.withStack>: while executing query to fetch Hub Implementation: All attempts fail:
  #1: graphql: 
```

Another time empty error from GraphQL... I think we should spend more time on debugging this in a dedicated issue.

PR #388, opened 2021-07-14
Hide stdout from shell commands

lukaszo:
Fix permissions for capact-local-ca.crt file on Linux systems

## Description

- Hide stdout brom shell commands
- Fix permissions for capact-local-ca.crt file on Linux systems
- 
## Related issue(s)

#328

### DIFF
--- a/internal/cli/capact/os.go
+++ b/internal/cli/capact/os.go
@@ -28,7 +28,7 @@ func AddGatewayToHostsFile(status *printer.Status) error {
 
 	status.Step("Updating /etc/hosts file")
 	// #nosec G204
-	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("echo \"%s\"| sudo tee -a /etc/hosts", entry))
+	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("echo \"%s\"| sudo tee -a /etc/hosts >/dev/null", entry))
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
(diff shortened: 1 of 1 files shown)


mszostok (approved):
still works on mac 👍

internal/cli/capact/os.go
```diff
@@ -69,7 +61,7 @@ func TrustSelfSigned(status *printer.Status) error {
 
 func trustSelfSignedDarwin(tmpCertPath string) error {
 	// #nosec G204
-	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("security verify-cert -c %s", tmpCertPath))
+	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("security verify-cert -c %s >/dev/null", tmpCertPath))
 	cmd.Stdout = os.Stdout
```
mszostok, line 65: isn't it easier to use io.Discard instead of having `/dev/null` in command?

mszostok:
it still doesn't look good, but it's acceptable:
```
 ✓ Installing capact Helm chart
 ⣯ Trusting self-signed CA certificate if not already trusted Cert Verify Result: CSSMERR_TP_NOT_TRUSTED
 ⣾ Trusting self-signed CA certificate if not already trusted
 ✓ Trusting self-signed CA certificate if not already trusted
```

internal/cli/capact/os.go
lukaszo: ah, in this case yes. In cases when I'm using sudo no

PR #389, opened 2021-07-15
Share Capact CLI between GH Action steps

mszostok:
## Description

- Share Capact CLI between GH Action steps

We use the `pull_request_target` event, so the `workflow` files are taken from the `main` branch, so my changes are not taken into account.

I tested that on my fork, see: https://github.com/mszostok/capact/runs/3076727952

## Related issue(s)

 - https://github.com/capactio/capact/issues/328

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -1,7 +1,7 @@
 name: PR build
 
 on:
-  pull_request:
+  pull_request_target:
     types: [ opened, synchronize, reopened ]
     paths:
       - "**.go"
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-build.yaml
```diff
@@ -272,3 +289,7 @@ jobs:
           name: cluster_dump_${{github.sha}}
           path: "output"
           retention-days: 5 # Default 90 days
+      - name: Delete Capact CLI artifact
```
pkosiec, line 292: 🤔 Hmm, maybe the artifact deletion is not necessary after all?
I think this is a nice feature to have built CLI for a given PR, similarly how we have the Docker images built. But it would be great to build binary also for Mac, which will increase the build time a little bit, but it would be great for PR review

So two options:
1.  Build also CLI binary for mac, do not delete artifacts, keep the `retention-days` low (e.g. 3 days? 🤔)
2. Do it exactly as it is right now

I'm fine with both approaches, I just wanted to highlight that it could be converted to nice feature for us.

pkosiec (approved):
LGTM

.github/workflows/pr-build.yaml
```diff
@@ -231,12 +231,18 @@ jobs:
         # We use a new `no_unique_dist_dir` feature, which is not yet released
         run: go install github.com/goreleaser/goreleaser@b53dbb89d02aa3673782f3d063d7d5039446baac
```
pkosiec, line 232: BTW as we have new GoReleaser release with the feature we require (v0.173.2), maybe we can use it instead of using a specific commit on main branch?

lukaszo (approved):

.github/workflows/pr-build.yaml
mszostok: For now, I will leave it as it is. Building CLI locally takes around 7sec. Finding the artifact on GH, downloading it, adding perms will take much longer, so personally I would never use it. Additionally, it's not compressed, so it's ~50mb.

.github/workflows/pr-build.yaml
mszostok: it' a different story, but I will update it here :)

PR #400, opened 2021-07-20
Do not hide stdout when sudo is used

lukaszo:
## Description

- fixes missing sudo password prompt


lukaszo:
Needs rethinking

Issue #399, opened 2021-07-19
Remove Type composition and signature usage in Hub manifests 


pkosiec:
## Description

AC:

- Remove Type composition usage in our all manifests [1SP]
    - clean up all manifests
    - for Interface, support `input/output.jsonSchema.ref`
- Remove unused/misleading signature property from our manifests
  - [x] manifests
  - [x] cypher in db populator
  - [x] documentation
  - [x] JSONSchema
  - [x] GraphQL Schema

## Reason

Unblock validation #398, remove misleading fields

PR #402, opened 2021-07-22
Add support for typeRef in Interface parameters

mszostok:
## Description

- Add support for typeRef in Interface parameters

## Testing
To run `capact validate` against real manifests, check https://github.com/capactio/hub-manifests/pull/11

## Related issue(s)

- https://github.com/capactio/capact/issues/399


ocf-spec/0.0.1/schema/common/json-schema-type.json
```diff
@@ -6,8 +6,7 @@
   "properties": {
     "value": {
       "type": "string",
-      "title": "Inline JSON Schema definition for the parameters. It may contain external references to JSON schemas in Types.",
-      "description": "Inline JSON Schema definition for the parameters."
+      "description": "Inline JSON Schema definition for the parameters. It may contain external references to JSON schemas in Types."
```
mszostok, line 9:
```suggestion
      "description": "Inline JSON Schema definition for the parameters."
```

ocf-spec/0.0.1/schema/interface.json
```diff
@@ -68,10 +68,25 @@
                 "type": "object",
                 "description": "The input parameters for a given Action.",
                 "properties": {
+                  "typeRef": {
+                    "$ref": "https://capact.io/schemas/common/type-ref.json"
+                  },
                   "jsonSchema": {
```
lukaszo, line 74: Why you haven't removed it?

lukaszo (approved):

lukaszo (approved):

ocf-spec/0.0.1/schema/interface.json
mszostok: This is for input parameters. In some cases we don't need to create dedicated type. Sometimes is better to inline jsonschema as we do not want to create TI, and we also don't reuse it anywhere, so to reduce the boilerplate you can just put directly in Interface manifest.

Issue #397, opened 2021-07-19
Migrate GKE and EKS installation scripts to use CLI


pkosiec:
## Description

- Migrate Capact installation in our tutorials:
    - AWS EKS installation
    - GCP GKE installation
- Migrate Capact installation in the [`recreate_cluster`](https://github.com/capactio/capact/blob/main/.github/workflows/recreate_cluster.yaml) CI job
    - Make sure it is tested (we can recreate the cluster if needed)

## Reason

We should have an unified way to install Capact on all environments without using Bash scripts for Capact components installlation anymore

PR #403, opened 2021-07-23
Adjust GKE and EKS tutorials and installation scripts to use capact install

Trojan295:
## Description

- adjust EKS deployment script for CLI based installation

## Related issue(s)

#397


hack/ci/cluster-components-install-upgrade.sh
```diff
@@ -24,7 +24,7 @@ main() {
     export DOCKER_TAG=${OVERRIDE_DOCKER_TAG:-${DOCKER_TAG}}
     export DOCKER_REPOSITORY=${OVERRIDE_DOCKER_REPOSITORY:-${DOCKER_REPOSITORY}}
     export PRINT_INSECURE_NOTES="false"
-    capact::install_upgrade::charts
+    capact::install
```
lukaszo, line 27: I've hard-coded  `--version=@local` in the `capact install` command. You would need to adjust it to set `--version=@latest --helm-repo-url=@latest` Then it will use the latest version from the main branch.

hack/ci/cluster-components-install-upgrade.sh
```diff
@@ -24,7 +24,7 @@ main() {
     export DOCKER_TAG=${OVERRIDE_DOCKER_TAG:-${DOCKER_TAG}}
     export DOCKER_REPOSITORY=${OVERRIDE_DOCKER_REPOSITORY:-${DOCKER_REPOSITORY}}
     export PRINT_INSECURE_NOTES="false"
```
mszostok, line 26: we still have `export CLUSTER_TYPE="GKE"` uppercase, we should unify that.

hack/eks/install.sh
```diff
@@ -75,7 +75,12 @@ capact::aws::install::efs_csi_driver() {
 
 capact::aws::install::capact() {
   shout "Deploying Capact..."
-  "${CURRENT_DIR}"/cluster-components-install-upgrade.sh
+  capact install --environment eks \
```
mszostok, line 78: Currently, `--environment eks` is not used, and it's quite inconsistent with ` --environment kind`.

Which approach we choose?
1. Set overrides via flags
2. Embed overrides in Capact CLI
3. Rename the `--environment` to sth kind related setup? BTW it's even not too much `kind` related setup but more `local` setup. With k3d we will apply the same overrides.

What I like about the current approach is that I can download and execute `capact install --env kind` and that's it. I do not have to download/set any overrides. On the other hand, embed overrides for eks, gke etc. may be hard to maintain/use.

@lukaszo @pkosiec any thoughts?

_NOTE:_
After final decision, get rid of `TODO` comments in Capact CLI: https://github.com/capactio/capact/search?l=Go&q=eks

hack/eks/install.sh
```diff
@@ -99,7 +104,7 @@ capact::aws::register_dnses() {
 
   local internal_lb_fqdn
   for _ in $(seq 6); do
-    internal_lb_fqdn="$(kubectl -n capact-system get svc ingress-nginx-controller '-ojsonpath={.status.loadBalancer.ingress[].hostname}')"
+    internal_lb_fqdn="$(kubectl -n capact-system get svc ingress-controller-ingress-nginx-controller '-ojsonpath={.status.loadBalancer.ingress[].hostname}')"
```
mszostok, line 107: you have outdated version of CLI, this release name was reverted.

hack/eks/install.sh
pkosiec: I like 3rd option the most. The second most preferred solution by me would be the No 2 one. However, I think the AWS-specific overrides are opinionated (e.g. whether the load balancer should be internal or not), so I think we don't need dedicated ones. So, we don't need `--environment aws` at all.

lukaszo: From what I see `--environment eks` is not used anywhere in CLI and only passed to not apply `kind` vars. Am I right?

In that case maybe we can just pass `--environment none` and pass all in flags?

mszostok: @lukaszo 
> is not used anywhere in CLI
yes, because instead of embedding, overrides are passed via flags. 

>In that case maybe we can just pass --environment none and pass all in flags?
This could work, but still we should change the `kind` env as overrides are not kind related. Additionally, we need to get rid of `TODOs` comments, plus probably get rid of `--env` flag at all as it may be misleading, we have currently `kind`, in `TODOs` we put support for `eks`, but now we have situation when for `kind` we have embedded overrides and enable them via `--env=kind` and for EKS, GKE, we set `--env=eks/gke` which skips the local related overrides and other overrides are passed via flags. For me, this inconsistency is bad, mostly because of UX.

internal/cli/capact/capact.go
```diff
@@ -13,8 +13,13 @@ const (
 	LocalDockerTag = "dev"
 	// LocalDockerPath path used when building local images
 	LocalDockerPath = "local"
+
 	// KindEnv default name for kind environment
 	KindEnv = "kind"
+	// GkeEnv default name for GKE environment
+	GkeEnv = "gke"
+	// EksEnv default name for EKS environment
+	EksEnv = "eks"
```
mszostok, line 22:
```suggestion
	// GKEEnv default name for GKE environment
	GKEEnv = "gke"
	// EKSEnv default name for EKS environment
	EKSEnv = "eks"
```
respectively the rest code needs to be updated

mszostok (approved):
I didn't test that, but introduced changes from the code perspective lgtm 👍 

Before merge, please recreate the long-running cluster using workflow from this branch 👍

PR #406, opened 2021-07-23
Add support for hub-manifest ref update, update ref to latest ver

mszostok:
Unfortunately, we cannot update HUB_MANIFESTS_SOURCE_REPO_REF as CI manifests are taken from main. Additionally, we use forks, so we are not able to test content from PRs only by changing commit SHA. This PR fixes this problem by allowing to override both `repo url` and `ref`.

## Description

- Add support for hub-manifest ref update, 
- Update hub-manifests ref to the latest version,
- Support optional sshKey. Currently, we do not need to have sshkey as hub-manifests repo is open-sourced.

# Testing

To test that optional the `sshKey` param works properly, run such commands:

```bash
# Test with `sshKey`
helm template --set hub-public.populator.manifestsLocation.sshKey='test' ./deploy/kubernetes/charts/capact | grep -m 1 -A 1 MANIFESTS_PATH

# Test without `sshKey`
helm template ./deploy/kubernetes/charts/capact | grep -m 1 -A 1 MANIFESTS_PATH
```

## Related issue(s)

- https://github.com/capactio/capact/issues/399

### DIFF
--- a/hack/lib/const.sh
+++ b/hack/lib/const.sh
@@ -46,3 +46,5 @@ readonly CAPACT_USE_TEST_SETUP="false"
 #
 
 readonly CAPACT_INCREASE_RESOURCE_LIMITS="true"
+readonly HUB_MANIFESTS_SOURCE_REPO_URL="git@github.com:capactio/hub-manifests.git"
+readonly HUB_MANIFESTS_SOURCE_REPO_REF="5464edbf0b3f5ef2d1a02c6266dfc50fb5ac0a4a"
--- a/hack/lib/const.sh
+++ b/hack/lib/const.sh
@@ -46,5 +46,5 @@ readonly CAPACT_USE_TEST_SETUP="false"
 #
 
 readonly CAPACT_INCREASE_RESOURCE_LIMITS="true"
-readonly HUB_MANIFESTS_SOURCE_REPO_URL="git@github.com:capactio/hub-manifests.git"
-readonly HUB_MANIFESTS_SOURCE_REPO_REF="5464edbf0b3f5ef2d1a02c6266dfc50fb5ac0a4a"
+readonly CAPACT_HUB_MANIFESTS_SOURCE_REPO_URL="github.com/capactio/hub-manifests"
+readonly CAPACT_HUB_MANIFESTS_SOURCE_REPO_REF="5464edbf0b3f5ef2d1a02c6266dfc50fb5ac0a4a"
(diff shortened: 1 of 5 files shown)


hack/lib/utilities.sh
```diff
@@ -218,6 +218,10 @@ capact::install() {
       CAPACT_OVERRIDES+=",hub-public.populator.manifestsLocation.branch=${HUB_MANIFESTS_SOURCE_REPO_REF}"
     fi
 
+    if [ -n "${HUB_MANIFESTS_SOURCE_REPO_URL:-}" ]; then
```
lukaszo, line 221: Please add `HUB_MANIFESTS_SOURCE_REPO_URL`  to the list of accepted variables at the top of the function.

With current approach we can not override is it always use the var from consts.sh is this expected?

mszostok: that's a good point, probably we should allow to override it. Changes introduced to `workflow` are not used so the `HUB_MANIFESTS_SOURCE_REPO_URL` will be still set for the old commit though.

mszostok:
Related PR: https://github.com/capactio/website/pull/46

hack/lib/const.sh
```diff
@@ -46,3 +46,5 @@ readonly CAPACT_USE_TEST_SETUP="false"
 #
 
 readonly CAPACT_INCREASE_RESOURCE_LIMITS="true"
+readonly CAPACT_HUB_MANIFESTS_SOURCE_REPO_URL="github.com/capactio/hub-manifests"
+readonly CAPACT_HUB_MANIFESTS_SOURCE_REPO_REF="5464edbf0b3f5ef2d1a02c6266dfc50fb5ac0a4a"
```
lukaszo, line 50: Let's add some comment here about how to get this value

mszostok: TBH I don't understand what kind of comment I can put here. It's mapped to those values:
https://github.com/capactio/capact/blob/78dc5c67e81ac6eba32e07a41a8c35d4239ffe81/deploy/kubernetes/charts/capact/charts/hub-public/values.yaml#L68-L71

I added description of a new flag in related PR: https://github.com/capactio/website/pull/46/files#diff-30642f79ee617e8a1dc888506e5cb08d4173a93d28f08006e1ffb5929cee644a

could you suggest a comment that I can put here? Thanks in advance.

lukaszo: I just want to add a comment how to get the value `5464edbf0b3f5ef2d1a02c6266dfc50fb5ac0a4a`  and when it needs to be updated.

lukaszo (approved):

PR #408, opened 2021-07-23
Increases Local hub resourcess during the upgrade

lukaszo:
## Description

- Increases Local hub resourcess during the upgrade

It was added for `install` command. Now also adding it for upgrade.

## Related issue(s)

#328


mszostok (approved):

internal/cli/upgrade/upgrade.go
```diff
@@ -205,6 +205,7 @@ func (u *Upgrade) resolveInputParameters(opts *Options) error {
 	if opts.Parameters.IncreaseResourceLimits {
```
mszostok, line 205: **nit**
maybe we should extract it to dedicated method and use in both places?

sth what we already have for `opts.Parameters.ResolveVersion()`

lukaszo: Yes, it will be in one of the follow-up tasks

Issue #411, opened 2021-07-26
Create/Use dedicated package for string slices


lukaszo:
## Description

Currently, in a different places of the code we have functions to, for example, check if slice [contains a string](https://github.com/capactio/capact/blob/main/internal/cli/capact/components.go#L560). 

The task is to find such duplicated pieces and move them to a one package. We may re-use https://github.com/kubernetes/utils/blob/master/strings/slices/slices.go or base our solution on it.

It's a good issue for a first-time contributor.

krishnaindani:
hey, could I try to work on this issue?

platten:
We welcome all contributions 😀 If you have any questions, please feel free to reach out to us on our [slack channel](https://capact.io/slack). Thanks!

krishnaindani:
I only see one place where this is used, so created a PR here https://github.com/capactio/capact/pull/416

PR #416, opened 2021-08-01
use strings slices from k8s utils

krishnaindani:
## Description

- Use slices functions for Kubernetes utils package

## Related issue(s)

Fixes https://github.com/capactio/capact/issues/411


lukaszo:
Hi, @krishnaindani thank you for your contribution. This part looks ok.

There should be similar check in the `build.go` file but it seems it's missing. Can you add it also there?https://github.com/capactio/capact/blob/main/internal/cli/capact/build.go#L125 You should iterate over `Images.All()` and check if `names` contain `name`

lukaszo (commented):
Hi, @krishnaindani thank you for your contribution. This part looks ok.

There should be similar check in the `build.go` file but it seems it's missing. Can you add it also there?https://github.com/capactio/capact/blob/main/internal/cli/capact/build.go#L125 You should iterate over `Images.All()` and check if `names` contain `name`

krishnaindani:
> Hi, @krishnaindani thank you for your contribution. This part looks ok.
> 
> There should be similar check in the `build.go` file but it seems it's missing. Can you add it also there?https://github.com/capactio/capact/blob/main/internal/cli/capact/build.go#L125 You should iterate over `Images.All()` and check if `names` contain `name`

sure, I will take a look at it.

krishnaindani:
@lukaszo I have updated that function, let me know how that looks.

internal/cli/capact/build.go
```diff
@@ -122,14 +123,14 @@ func DeleteImages(images []string) error {
 func BuildImages(w io.Writer, repository, version string, names []string) ([]string, error) {
 	var created []string
 
-	for _, name := range names {
-		image, ok := Images[name]
+	for _, image := range Images.All() {
+		ok := slices.Contains(names, image)
 		if !ok {
-			return nil, fmt.Errorf("cannot find image %s", name)
+			return nil, fmt.Errorf("cannot find image %s", image)
```
lukaszo, line 129: if `image` is not in the `names` just continue. Don't return error

lukaszo (approved):
Looks good. Thanks :+1:  :)

krishnaindani:
Thank you

PR #422, opened 2021-08-03
Add Interface and Terraform Implementation content creator

Trojan295:
## Description

- Add `capact alpha content interface` command to bootstrap Interface manifests,
- add `capact alpha content terraform` command to bootstrap Terraform-based Implementation manifests,
- export a public function to get manifest metadata.

## Improvment options

- Fill the parameters in the `fill-default-input` step in Terraform-based Implementation, based on the input-parameters read from the Interface. This would require to parse the JSON schema + if we want to read the Interface from filesystem (not Hub), then we have to prepare a custom loader, cause on GraphQL the manifest syntax is a bit different,
- subcommand to create a tarball with the Terraform module. Modules in general don't have a `provider` resource inside it (e.g. [like this](https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-user)), so we could include it and expose some additional variables to configure it before packing the tarball.

Recognizion of how to wire Terraform inputs/outputs to Interface inputs/outputs can be hard, as the names of this properties could not match always. E.g. we could have a Interface input parameter `name`, which would be mapped to the Terraform variable `prefix`. Thus Content Creators have to decide, which variables are from Interface parameters and which from Implementation additional parameters. They also have to build the final output TypeInstance structure.
**This would be an awesome case to use AI.**

## Testing

Best way IMO would be to try to create manifests for a Terraform module, so you can get a feeling on how this improves the content creation. I created a tarball with [this AWS IAM user module](https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-user) (with small adjustments to it).

```bash
go run cmd/cli/main.go alpha manifest-gen interface cap.interface.cloud.user.create
go run cmd/cli/main.go alpha manifest-gen implementation terraform cap.implementation.aws.iam.user.create ../terraform-aws-iam/modules/iam-user -i cap.interface.cloud.user.create -s "https://storage.googleapis.com/capactio-terraform-modules/terraform.aws.iam.user/0.1.0/module.tgz" -p aws
```

This will generate to you the manifests, which you have to fill the credentals TypeInstance, provide the AWS keys to Terraform step and set proper inputs and outputs.


cmd/cli/cmd/alpha/content/content.go
```diff
@@ -0,0 +1,21 @@
+package content
+
+import "github.com/spf13/cobra"
+
+var manifestOutputDirectory string
+
+// NewCmd returns a cobra.Command for content generation operations.
+func NewCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "content",
```
pkosiec, line 10: I think we should put "generator/gen/generate" somewhere in the command name. Maybe `capact alpha manifest-gen ...`?

cmd/cli/cmd/alpha/content/interface.go
```diff
@@ -0,0 +1,42 @@
...
+	"capact.io/capact/internal/cli/heredoc"
+	"github.com/spf13/cobra"
+)
+
+var interfaceCfg content.InterfaceConfig
+
+// NewInterface returns a cobra.Command to bootstrap new Interface manifests.
+func NewInterface() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "interface [PREFIX] [NAME]",
```
pkosiec, line 17: I think splitting prefix and name could be confusing for users. Maybe the user should provide a full path, even including `cap.interface.`?

E.g.:

```
capact alpha manifest-gen interface cap.interface.database.postgresql.install
```

cmd/cli/cmd/alpha/content/interface.go
```diff
@@ -0,0 +1,42 @@
...
+		Long:  "Bootstrap new Interface and associated Type manifests",
+		Example: heredoc.WithCLIName(`
+			# Bootstrap manifests for the cap.interface.database.postgresql.install Interface
+			<cli> alpha content interface database.postgresql install`, cli.Name),
+		Args: cobra.ExactArgs(2),
+		Run: func(cmd *cobra.Command, args []string) {
+			interfaceCfg.ManifestsPrefix = args[0]
+			interfaceCfg.ManifestName = args[1]
+
+			files, err := content.GenerateInterfaceManifests(&interfaceCfg)
```
pkosiec, line 28: Honestly, I thought it would ask me about some parameters, e.g. some metadata fields or at least input/output of the Interface 😞

cmd/cli/cmd/alpha/content/content.go
```diff
@@ -0,0 +1,21 @@
...
+		Short: "Content generation",
+		Long:  "Subcommand for various content generation operations",
+	}
+
+	cmd.AddCommand(NewInterface())
+	cmd.AddCommand(NewTerraform())
+
```
pkosiec, line 17: We also need to generate InterfaceGroup. It could be a separate command or we can add a flag to generate it along Interface.

internal/cli/alpha/content/terraform.go
```diff
@@ -0,0 +1,368 @@
...
+		return "boolean"
+	case "null":
+		return "null"
+	}
+
+	return "object"
+}
+
+const (
+	terraformImplementationManifestTemplate = `ocfVersion: 0.0.1
```
pkosiec, line 150: I think it would be much more readable if we move the manifest templates out of Go code to separate files.

internal/cli/alpha/content/terraform.go
```diff
@@ -0,0 +1,368 @@
...
+        path: "cap.type.{{ .Prefix }}.{{ .Name }}-input"
+        revision: 0.1.0
+
+  outputTypeInstanceRelations:
+    config:
+      uses:
+        - terraform-release
+
+  implements:
+    - path: {{if .InterfacePath}}cap.interface.{{ .InterfacePath }}{{else}}"cap.interface..." # Put here the path of the implemented Interface{{end}}
```
pkosiec, line 181: It would be easier to find places which user should edit if you add `TODO:` prefix to all your comments that require action from user.

cmd/cli/cmd/alpha/content/helpers.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func writeManifestFiles(files map[string]string) error {
```
pkosiec, line 12: I think it would be better to fail if files in a given path already exist. Someone can override already generated and adjusted manifests with new command run.

BTW And please move it out of this directory into `internal`. Let's have just files which are related to commands only.

cmd/cli/cmd/alpha/content/terraform.go
```diff
@@ -0,0 +1,52 @@
+package content
```
pkosiec, line 1: The interactive mode would be useful - but IMHO it can be implemented in a follow-up.

cmd/cli/cmd/alpha/content/terraform.go
```diff
@@ -0,0 +1,52 @@
...
+	"capact.io/capact/internal/cli"
+	"capact.io/capact/internal/cli/alpha/content"
+	"capact.io/capact/internal/cli/heredoc"
+	"github.com/spf13/cobra"
+)
+
+var tfContentCfg content.TerraformConfig
+
+// NewTerraform returns a cobra.Command to bootstrap Terraform based manifests.
+func NewTerraform() *cobra.Command {
```
pkosiec, line 15: I'm wondering if it's good to generate the Terraform values as additionalInput, instead of input parameters for Interface 🤔  But then of course we would need to generate Interface along Terraform Implementation, to avoid calling Hub API.

pkosiec (commented):
Generally good work 👍 

To discuss with team in which direction we want to go:
1. What do we do with Terraform input - currently the inputs are generated as `additionalInput`. I'm not sure if this is a good behavior.
1. InterfaceGroup generation - I think we should implement it in this PR. The question is whether we go with separate command, or we generate InterfaceGroup along with Interface for `terraform` command
1. Add ability to provide more values during generation (e.g. for metadata - the fields are usually the same/similar across Interface, InterfaceGroup, Types, Implementation). I think it would be great to implement it now, but maybe others have different opinion on that
1. Interactive mode (should we implement it within this task? IMHO we should postpone it to a follow-up)

WDYT @mszostok @lukaszo?

cmd/cli/cmd/alpha/content/terraform.go
Trojan295: I wanted to split the Interface generation and Implementation generation, as you could already have an Interface and you just want to generate an Implementation for it. Also in most of our cases the Terraform variables end up in the additionalParameters.

cmd/cli/cmd/alpha/content/content.go
Trojan295: Yeah, I forgot about it

internal/cli/alpha/content/terraform.go
Trojan295: I thought about it, but it could be a bit misleading, if this TODO is code related or for the Content Creator.

mszostok:
> What do we do with Terraform input - currently the inputs are generated as additionalInput. I'm not sure if this is a good behavior.

I'm ok that it's generated as `additionalInput` 👍 

>  InterfaceGroup generation - I think we should implement it in this PR. The question is whether we go with separate command, or we generate InterfaceGroup along with Interface for terraform command

👍 

>  Add ability to provide more values during generation (e.g. for metadata - the fields are usually the same/similar across Interface, InterfaceGroup, Types, Implementation). I think it would be great to implement it now, but maybe others have different opinion on that

not sure about that one but would be nice to have 👍 

>  Interactive mode (should we implement it within this task? IMHO we should postpone it to a follow-up)

IMO follow-up is good. Current version is mostly for us and IMO interactive mode is good for demos/users who try this for the first time. From my experience, I quite often avoid interactive mode as it's slowing me down.


---
regarding command structure, I like it a lot 👍  it shows the Capact vision that Interfaces are different part than Implementation. First you need to have Interface, next you can generate Implementation. Doing that in single step (e.g. as a part of Impl generation) could mean that for each implementation is good to have Interface which is wrong assumption. So +1 for this split.

But I would vote to change one thing:

- `capact alpha content interface create cap.foo` -  bootstrap Interface manifests - its ok 👍 
- `capact alpha content implementaiton create cap.foo`  - bootstrap generic Implementation manifests
- `capact alpha content implementaiton create cap.foo --for=terraform` - bootstrap Terraform aware Implementation
- `capact alpha content implementaiton create cap.foo --for=helm` - etc.

It doesn't need to be named `--for` but I just want to show the idea here.

Trojan295:
One drawback I see in `--for` is that if you want to add additional flags related to only one of the specific Implementations (e.g. provider for Terraform, which is already), then you have to add it to the generic `capact alpha content implementaiton create cap.foo` cmd, right?

mszostok:
@Trojan295 that's a good point. Provider IMO can be implemented also for generic as `requires` needs to be always and for generic we can have just a small example step in workflow or even comment that `workflow.outputs.artifacts.gcp-credentials` will be injected and can be used by you.

But still there may be some impl specific flag and it could cause problem. We can have a validator and pin it to a specific Impl. Currently cobra doesn't support that: https://github.com/spf13/cobra/issues/413

So maybe sth like:

```
capact alpha content impl generic create cap.foo
capact alpha content impl terraform create cap.foo
```

cmd/cli/cmd/alpha/content/helpers.go
Trojan295: I added a `--override` flag, which is set to `false` by default.

pkosiec (commented):
Few comments related to code 🙂 
Disclaimer: I still need to run it and test the manifest running.

internal/cli/alpha/manifestgen/templates.go
```diff
@@ -0,0 +1,340 @@
...
+        path: "cap.type.{{ .Prefix }}.{{ .Name }}-input"
+        revision: 0.1.0
+
+  outputTypeInstanceRelations:
+    config:
+      uses:
+        - terraform-release
+
+  implements:
+    - path: {{if .InterfacePath}}cap.interface.{{ .InterfacePath }}{{else}}"cap.interface..." # Put here the path of the implemented Interface{{end}}
```
pkosiec, line 153: Using full paths everywhere will be more consistent.

```suggestion
    - path: {{if .InterfacePath}}{{ .InterfacePath }}{{else}}"cap.interface..." # Put here the path of the implemented Interface{{end}}
```

I ran:

```
go run cmd/cli/main.go alpha manifest-gen implementation terraform cap.implementation.aws.iam.user.create ../terraform-aws-iam/modules/iam-user -i cap.interface.cloud.user.create -s "https://storage.googleapis.com/capactio-terraform-modules/terraform.aws.iam.user/0.1.0/module.tgz" -p aws
```

internal/cli/alpha/manifestgen/templates.go
```diff
@@ -0,0 +1,340 @@
...
+                  capact-action: jinja2.template
+                  arguments:
+                    artifacts:
+                      - name: template
+                        from: "{{"{{"}}steps.fill-parameters.outputs.artifacts.render{{"}}"}}"
+                      - name: input-parameters
+                        {{if eq .Provider "aws" -}}
+                        from: "{{"{{"}}workflow.outputs.artifacts.aws-credentials{{"}}"}}"
+                        {{- else if eq .Provider "gcp" -}}
+                        from: "{{"{{"}}workflow.outputs.artifacts.gcp-credentials{{"}}"}}"
```
pkosiec, line 280:
```suggestion
                        from: "{{"{{"}}workflow.outputs.artifacts.gcp-sa{{"}}"}}"
```

pkosiec (commented):
A last batch of comments after using the CLI after changes. I validated these manifests and the result is `🚀 No errors detected.`.

Did you try to run the manifests?

internal/cli/alpha/manifestgen/templates.go
```diff
@@ -0,0 +1,340 @@
+package manifestgen
```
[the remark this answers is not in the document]
Trojan295: I extracted the templates and used `embed` package to bake them into the binary at compile time.

internal/cli/alpha/manifestgen/provider.go
```diff
@@ -0,0 +1,37 @@
+package manifestgen
```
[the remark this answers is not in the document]
Trojan295: invalid after discussions

pkosiec (approved):
LGTM 🚀 Great work!

PR #430, opened 2021-08-06
Allow running PR builds from untrusted contributors

Trojan295:
## Description

- split build and pushing images for apps, tests and infra to two separate jobs:
  - build job has only read permissions
  - push job has write permissions, but uses the scripts from main
- add Make target to export images a tarballs
- remove some `.PHONY` targets as they don't work with pattern rules

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -168,6 +168,7 @@ jobs:
     permissions:
       contents: read
       packages: write
+    steps:
       - name: Checkout code
         uses: actions/checkout@v2
       - name: Download artifact
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -154,7 +154,7 @@ jobs:
       - name: Upload artifact
         uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.APP }}-${{ github.event.pull_request.number }}
+          name: ${{ matrix.APP }}-${{github.sha}}
           path: /tmp/${{ matrix.APP }}.tar
           retention-days: 5
 
@@ -171,10 +171,12 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+        with:
+          persist-credentials: false
       - name: Download artifact
         uses: actions/download-artifact@v2
         with:
-          name: ${{ matrix.APP }}-${{ github.event.pull_request.number }}
+          name: ${{ matrix.APP }}-${{github.sha}}
           path: /tmp
       - name: Load image
         run: docker load --input /tmp/${{ matrix.APP }}.tar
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-build.yaml
```diff
@@ -1,7 +1,7 @@
 name: PR build
 
 on:
-  pull_request_target:
+  pull_request:
```
Trojan295, line 4: This is put just for demo purpose, I will change this back to `pull_request_target` before merging

.github/workflows/pr-build.yaml
```diff
@@ -149,9 +148,45 @@ jobs:
...
+          make build-app-image-${{ matrix.APP }}
+          make save-app-image-${{ matrix.APP }}
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
+          path: /tmp/${{ matrix.APP }}.tar
+          retention-days: 1
+
+  push-apps:
```
mszostok, line 161: I thought about sharing a common part, e.g.
```yaml
  push: &genericmetapush
    runs-on: ubuntu-latest
    if: github.event.pull_request.draft == false
    permissions:
      contents: read
      packages: write
  steps: &genericsteps
    - name: Checkout code
      uses: actions/checkout@v2
      with:
        persist-credentials: false
    - name: Download artifact
      uses: actions/download-artifact@v2
      with:
        name: ${{ matrix.APP }}-${{github.sha}}
        path: /tmp
    - name: Log into registry
      run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
    - name: Setup environment
      env:
        PR_NUMBER: ${{ github.event.pull_request.number }}
      run: |
        . ./hack/ci/setup-env.sh

  push-apps: *genericmetapush
    name: Push ${{ matrix.APP }}
    needs: [ prepare-matrix, build-apps ]
    strategy:
      matrix: ${{fromJson(needs.prepare-matrix.outputs.matrix-app)}}
    steps: *genericsteps
    - name: Load image
      run: docker load --input /tmp/${{ matrix.APP }}.tar
    - run: make push-app-image-${{ matrix.APP }}
```

but after trying that I see that it decreases readability and at the same time may cause some security issues.

mszostok (approved):
lgtm 👍

.github/workflows/pr-build.yaml
```diff
@@ -149,9 +148,45 @@ jobs:
...
+      - name: Load image
+        run: docker load --input /tmp/${{ matrix.APP }}.tar
       - name: Log into registry
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+      - name: Setup environment
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          . ./hack/ci/setup-env.sh
       - run: make push-app-image-${{ matrix.APP }}
```
mszostok, line 190: consider adding delete step after pushing
```yaml
      - name: Delete Capact CLI artifact
        uses: geekyeggo/delete-artifact@v1
        with:
          name: ${{ matrix.APP }}-${{github.sha}}
```
I do not see reason why we want to keep it. Currently, we quite often push and build images, so it will be good to GC it earlier than  1day.

PR #440, opened 2021-08-10
Adjust populator and Hub GraphQL schema to support typeRef in Interface input-parameters

Trojan295:
## Description

- Adjust populator and Hub GraphQL schema to support typeRef in Interface input-parameters

You might need to use this manifest branch https://github.com/capactio/hub-manifests/pull/18 to test this.

### DIFF
--- a/hub-js/graphql/public/schema.graphql
+++ b/hub-js/graphql/public/schema.graphql
@@ -235,7 +235,8 @@ type InterfaceInput @additionalLabels(labels: ["published"]){
 
 type InputParameter @additionalLabels(labels: ["published"]){
   name: String!
-  typeRef: TypeReference! @relation(name: "OF_TYPE", direction: "OUT")
+  jsonSchema: Any 
+  typeRef: TypeReference @relation(name: "OF_TYPE", direction: "OUT")
 }
 
 type InterfaceOutput @additionalLabels(labels: ["published"]){
--- a/hub-js/graphql/public/schema.graphql
+++ b/hub-js/graphql/public/schema.graphql
@@ -235,7 +235,7 @@ type InterfaceInput @additionalLabels(labels: ["published"]){
 
 type InputParameter @additionalLabels(labels: ["published"]){
   name: String!
-  jsonSchema: Any
+  typeRef: TypeReference! @relation(name: "OF_TYPE", direction: "OUT")
 }
 
 type InterfaceOutput @additionalLabels(labels: ["published"]){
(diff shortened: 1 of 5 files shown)


pkg/hub/client/public/fields.go
```diff
@@ -137,7 +137,10 @@ var InterfaceRevisionFields = fmt.Sprintf(`
         input {
           parameters {
             name
-            jsonSchema
+						typeRef {
```
mszostok, line 140: please fix padding here :)

hub-js/graphql/public/schema.graphql
```diff
@@ -235,7 +235,7 @@ type InterfaceInput @additionalLabels(labels: ["published"]){
 
 type InputParameter @additionalLabels(labels: ["published"]){
   name: String!
-  jsonSchema: Any
+  typeRef: TypeReference! @relation(name: "OF_TYPE", direction: "OUT")
```
mszostok, line 238: The `jsonSchema` cannot be removed

see:
- https://github.com/capactio/capact/pull/402#discussion_r675419435
- https://structsure.slack.com/archives/C01DTR12JUA/p1627031890062100?thread_ts=1626961877.057900&cid=C01DTR12JUA

TL;DR; we already use it and thanks to that we also reduce the boilerplate.

pkg/sdk/dbpopulator/populator.go
```diff
@@ -137,9 +137,14 @@ CALL {
   WITH input, parameters
   UNWIND keys(parameters) as name
     CREATE (parameter: InputParameter:unpublished {
```
mszostok, line 139: only one thing can be specified, either `jsonSchema` or `typeRef`. You need to check which one was specified and populate a proper data.

mszostok:
FYI: 
To test changes made on Hub manifests, you can always use your fork on PR by changing those values: https://github.com/capactio/capact/blob/99ebe46f75411ed65f385cfb9eda87e60a6e0b4d/hack/lib/const.sh#L49-L52

In your case, it will be:
```bash
 readonly CAPACT_HUB_MANIFESTS_SOURCE_REPO_URL="github.com/Trojan295/hub-manifests" 
 # The git ref to checkout. It can point to a commit SHA, a branch name, or a tag. 
 # If you want to use your forked version, remember to update CAPACT_HUB_MANIFESTS_SOURCE_REPO_URL  respectively. 
 readonly CAPACT_HUB_MANIFESTS_SOURCE_REPO_REF="fix-jsonschema-manifest"

mszostok (approved):

Issue #415, opened 2021-07-29
`capact ti get` may cause gateway to be oomkilled


lukaszo:
## Description

`capact ti get` may cause gateway to be oomkilled. We pull all TypeInstance informations from the API. After running few actions I've got almost 30MB of a raw yaml. 

When this happens, user cannot see any data:

```shell
$ capact ti get
Error: while executing query to list TypeInstances: All attempts fail:
#1: decoding response: invalid character '<' looking for beginning of value
```

Ingress is returning 
```
2021/07/29 13:04:56 >> headers: map[Accept:[application/json; charset=utf-8] Content-Type:[application/json; charset=utf-8]]
2021/07/29 13:05:00 << <html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx</center>
</body>
</html>
```
Because Gateway was killed.

There are few options to solve it:
1. Short term: Increase gateway resources so it's not oomkilled
2. Mid term: Change CLI to query only required data. Fot table view it's only few fields. Problem persists when yaml output is used
3. Long term: All above plus something like pagination so we don't download all

### AC
- Increase gateway resources so it's not oomkilled
- Change CLI to query only required data. Fot table view it's only few fields. Problem persists when yaml output is used

mszostok:
4. It will be also good to think about some cache layer.

mszostok:
**Testing**
I run such test to verify if Gateway component is `OOMKilled`.

1. Create 8 TypeInstances with 1MB payload under `spec.value` via `capact ti stress -r 8 -s 1` (see https://github.com/capactio/capact/pull/443).
2. Build CLI from the latest main (https://github.com/capactio/capact/commit/608679fdce7c449adf7aa82f0b67356fd2c29c25)
3. Execute `capact ti get` result
   ```
   RoundTrip time: 31.028609749s Request size: 3536 Response byte size: 32799137
   ```
  So I was able to fetch 32MB, I also was able to fetch 41MB after adding more TI. The problem was that I had to manually change our HTTP timeout as for 41MB it took 1m5s to download all this data.

[...]

PR #443, opened 2021-08-12
Add option to select requested fields

mszostok:
## Description

- Add option to select requested fields
* Reduced:
  * For `table` requests we fetch only the required fields which in most cases are root fields (path, name, prefix) and metadata. Thanks to that, we reduced not only response body size but also request body size, e.g in case of `capact ti get`
    * Request size from `3536bytes` to `419bytes`
    * Response size from `32.79mb` to `4kb`
    * Request time from `31s` to `537ms`
* Print warning about timeout when getting all TI with all fields
* Remove not used filter from `func (c \*Client) ListImplementationRevisions(ctx context.Context, filter \*gqlpublicapi.ImplementationRevisionFilter) (\[\]\*gqlpublicapi.ImplementationRevision, error)`

## Next phase
  
  * Add Templates, and modify struct to have info under which property it should be attached. So thinks like`UsedByIDField` and `UsedByNameField` are not exclusive but can be set together. Currently, we don’t have such problem, but in the future it may be problematic. This will be good to implement too. We can use Go template for that.

 * more ideas: https://github.com/capactio/capact/issues/415#issuecomment-898305763

## Testing

First, create a cluster via `make dev-cluster`.

1. Cluster after installation:

* `capact ti get`
  * **\[main\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 1.469902062s Request size: 3536 Response byte size: 25713
  * **\[pr-443\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 192.565907ms Request size: 419 Response byte size: 2423

2\. RUN: `capact ti stress create -repeat 8 --payload-size 1`

* `capact ti get`
  * **\[main\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 31.028609749s Request size: 3536 Response byte size: 32799137
  * **\[pr-443\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 537.600118ms Request size: 419 Response byte size: 3983
* `capact hub implementations get`
  * **\[main\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 2.447752345s Request size: 2317 Response byte size: 169463
  * **\[pr-443\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 865.982963ms Request size: 771 Response byte size: 20122
* `capact hub interfaces get`
  * **\[main\]** RoundTrip time: 3.79519272s Request size: 3289 Response byte size: 186397
  * **\[pr-443\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 797.799794ms Request size: 1229 Response byte size: 31315
* `capact ti get -oyaml`
  * **\[main\]** First I got only error: `Error: while executing query to list TypeInstances: All attempts fail: #1: reading body: net/http: request canceled (Client.Timeout or context cancellation while reading body)`
  * **\[main\]** After retry: RoundTrip time[:](https://gateway.capact.local/graphql:) 31.398347837s Request size: 3536 Response byte size: 32799137
  * **via** `capact ti get -oyaml --timeout 90s`

    **\[pr-443\]** RoundTrip time[:](https://gateway.capact.local/graphql:) 25.596413962s Request size: 3570 Response byte size: 32799137
* Generic test via hyperfine for `capact ti get`:

  `hyperfine --export-markdown /tmp/results.md -L capact capact,./bin/capact-darwin-amd64 '{capact} ti get > /dev/null'`

    | Command | Mean \[s\] | Min \[s\] | Max \[s\] | Relative |
    |:---|---:|---:|---:|---:|
    | `capact ti get > /dev/null` | 36.796 ± 4.745 | 28.211 | 47.146 | 81.87 ± 92.20 |
    | `./bin/capact-darwin-amd64 ti get > /dev/null` | 0.449 ± 0.503 | 0.274 | 1.880 | 1.00 |

  **Summary** `'./bin/capact-darwin-amd64 ti get > /dev/null'` ran 81.87 ± 92.20 times faster than `'capact ti get > /dev/null'`

3. To clean-up create TI, run: `capact ti stress clean-up`
4. I tested diff between output new and old cli manually. 
5. Commands to get a single object were also fixed, e.g. get two type instances in table format `capact ti get 123 567` is also faster, same applies to interfaces and impls 

## Related issue(s)

- #415


pkg/hub/client/public/impl_fields.go
```diff
@@ -0,0 +1,100 @@
...
+
+// ifaceRevisionFieldsRegistry holds possible fields configuration for ImplementationRevision query.
+var implRevisionFieldsRegistry = map[ImplementationRevisionQueryFields]string{
+	ImplRevRootFields: `
+		revision`,
+	ImplRevMetadataFields: implRevisionMetadataFields,
+	ImplRevAllFields:      implRevisionAllFields,
+}
+
+var implRevisionMetadataFields = fmt.Sprintf(`
```
mszostok, line 13: lines 13-10 were not added but moved from: https://github.com/capactio/capact/pull/443/files#diff-d30b8a24c6a480408c406de4584ea3854c4571d3b6d609b22ba4fb9068d6c73aL32

pkg/hub/client/public/fields.go
```diff
@@ -19,7 +19,7 @@ var GenericMetadataFields = `
       iconURL
       `
 
-// AttributeFields for quering the Attributes fields and GenericMetadata.
+// AttributeFields for querying the Attributes fields and GenericMetadata.
 var AttributeFields = fmt.Sprintf(`
```
pkosiec, line 23: BTW I think we can unexport the `AttributeFields` and `GenericMetadataFields`, as they are not used anywhere else than in this package (and IMHO they shouldn't).

pkg/hub/client/public/iface_options.go
```diff
@@ -1,24 +1,41 @@
...
 	// Apply overrides
 	for _, opt := range opts {
 		opt(o)
 	}
 }
 
-// FindInterfaceRevisionOption provides an option to configure the find request for Interface Revision.
-type FindInterfaceRevisionOption func(*FindInterfaceRevisionOptions)
+// WithLatestIfaceRevision adds latestRevision fields for Interface query.
+func WithLatestIfaceRevision(requestedFields InterfaceRevisionQueryFields) InterfaceOption {
```
pkosiec, line 27: Honestly, I'm not a big fan of `Iface` in public func/method names 😄  I would suggest changing that to Interface, but it's up to you 🙂   Then we would need to rename the `WithImplRevCustomFields` option as well, which will be really, really long...

cmd/cli/cmd/typeinstance/get.go
```diff
@@ -158,3 +176,9 @@ func toTypeInstanceIDs(in []*gqllocalapi.TypeInstance) string {
 	}
 	return strings.Join(out, ", ")
 }
+
+func printHugePayloadWarning() {
+	warning := color.New(color.FgYellow).FprintfFunc()
+	warning(os.Stderr, "  Warning: ")
+	fmt.Fprintln(os.Stderr, "Fetching all TypeInstances with all fields. This may generate huge payload and request may be canceled due to timeout. Change '--timeout' value if needed.")
```
pkosiec, line 183:
```suggestion
	fmt.Fprintln(os.Stderr, "Fetching full details of all TypeInstances. This may take a while and generate huge output. Use '--timeout' flag to increase timeout if needed.")
```

pkosiec (approved):
LGTM, good work! I also think the bitmasking idea is the best option out there for configuring fields for GraphQL queries. 
I left just a few minor comments.


BTW:
1. While I ran `capact ti get -oyaml --timeout=3m` once I have lot of TypeInstances I got few times after one minute:

```bash
capact ti get -oyaml --timeout=3m

  Warning: Fetching all TypeInstances with all fields. This may generate huge payload and request may be canceled due to timeout. Change '--timeout' value if needed.
Error: while executing query to list TypeInstances: All attempts fail:
#1: decoding response: invalid character '<' looking for beginning of value
```

But the Gateway and public Hub Pods didn't restart. No logs for both of the Pods. Unfortunately I didn't install monitoring 😞 We need to monitor this issue and maybe do a follow-up if that happens often. Looks like this is something Łukasz initially observed (Bad gateway).

2. during `capact ti get -oyaml` I got panic, but it was due to the additional debug data in `httputil.configurableTransport.RoundTrip()`. After deleting it everything was fine.

pkg/hub/client/public/iface_options.go
mszostok: ok, for public types/funcs I used the InterfaceRevision and ImplementationRevision form 👍 

I also removed the `custom` from the the `WithImplRevCustomFields` as it redundant.

pkg/hub/client/local/options.go
```diff
@@ -0,0 +1,71 @@
...
+		opts.fields = getTypeInstanceFieldsFromFlags(queryFields)
+	}
+}
+
+func getTypeInstanceFieldsFromFlags(queryFields TypeInstancesQueryFields) string {
+	var names []string
+	for key := TypeInstanceRootFields; key < typeInstanceMaxKey; key <<= 1 {
+		if queryFields.Has(key) {
+			names = append(names, typeInstancesFieldsRegistry[key])
+		}
```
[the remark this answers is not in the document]
mszostok: it that will be an error handling, or there will be more logic after/or in this `if` then I would definitely change it. But for now it's concise and readable so I will leave it as it is 👍

pkosiec:
BTW I debugged the issue from point no 1., and the response I got is:
```
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

Interestingly enough, when I connected directly to local Hub and did the same command, I got connection closed by the server without any response:

```
2021/08/17 11:13:13 >> headers: map[Accept:[application/json; charset=utf-8] Content-Type:[application/json; charset=utf-8]]
Error: while executing query to list TypeInstances: All attempts fail:
#1: Post "http://localhost:3000/graphql": EOF
```

Again, no logs on Gateway and local Hub. I think we should debug it further in a follow-up. Personally I would start from seeing why Hub closes connections before returning a reply and add additional debug logging there for future.
Then we can see if Gateway reacts properly on edge cases. [From what I checked](https://github.com/nautilus/gateway/blob/master/http.go#L398) Nautilus Gateway responds with the following HTTP status codes:
```
200, 422, 501 i 400
```

Maybe [501](https://github.com/nautilus/gateway/blob/master/http.go#L162) results in a bad gateway error on nginx-ingress? Could be the case, as it happens when the response from underlying server cannot be serialized.

cmd/cli/cmd/typeinstance/get.go
```diff
@@ -17,12 +22,16 @@ import (
...
+const (
+	yamlFileSeparator   = "---"
+	tableRequiredFields = local.TypeInstanceRootFields | local.TypeInstanceTypeRefFields | local.TypeInstanceUsedByIDField | local.TypeInstanceUsesIDField | local.TypeInstanceLatestResourceVersionField
+)
 
 // GetOptions is used to store the configuration flags for the Get command.
 type GetOptions struct {
 	RequestedTypeInstancesIDs []string
 	ExportToUpdateFormat      bool
+	Timeout                   time.Duration
```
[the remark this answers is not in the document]
mszostok: fixed

pkosiec (approved):

Issue #417, opened 2021-08-02
Create content development tools for Helm charts


pkosiec:
## Description

Create content development tools for Helm charts

- Option 1: Do it as a PoC , write it in Go as a part of CLI. Create follow-up task if time ran out (e..g. to write missing unit tests, add additional features)
- Option 2: Do it as a PoC as fast as possible (in any language), we can run Docker container via our CLI

## Reason

We need to provide tools to quickly create Capact manifests for common use cases.

PR #445, opened 2021-08-16
Fix manifest generation

Trojan295:
## Description

- Fix manifest generation
- Add smoke tests for manifests generation to `make test-unit`

### DIFF
--- a/internal/cli/alpha/manifestgen/metadata.go
+++ b/internal/cli/alpha/manifestgen/metadata.go
@@ -7,7 +7,7 @@ import (
 
 // Metadata holds generic metadata information for Capact manifests.
 type Metadata struct {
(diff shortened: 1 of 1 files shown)


hack/test-unit.sh
```diff
@@ -63,12 +63,28 @@ function test::unit() {
   fi
 }
 
+function test::cli_manifest_gen() {
```
pkosiec, line 66: I think it's not really fits to the `test-unit` target. It's purely about executing command without cleanup.

It would be great to rewrite that in Go as a simple smoke test (comparing the output files to some golden files, similarly what we have for rendering). maybe as a part of #417 if we have time. For now I would delete that, honestly. What do you think?

pkosiec (approved):

PR #447, opened 2021-08-18
Add Helm manifest gen

Trojan295:
## Description

- Add `capact alpha manifest-gen implementation helm` command
- Rename `--override` flag to `--overwrite`
- Add unit tests for the manifest generation code

### Testing

You can use the following command to generate manifests for the `bitnami/postgresql` Helm chart:
```bash
go run cmd/cli/main.go alpha manifest-gen implementation helm cap.implementation.bitnami.postgresql.install postgresql --repo "https://charts.bitnami.com/bitnami" -i cap.interface.database.postgresql.install
```
### Problems

1. If user provides an empty parameter value for a string, it will be interpreted as null, when evaluating values in workflow.
   The values of the parameters are transformed to strings in format:
   ```yaml
   key: <@ additionalInput.key | default("some_default") @>
   # becomes, if user provides an empty string as value
   key:        # interpreted as null in YAML
   ```
   I couldn't find a way to make `yaml.Marshal` add double quotes to every string value so it would be 

2. If a key in Helm values.yaml contains a dot, then Jinja will have problem with interpreting it correctly. I'm not sure if this is a problem in Jinja as changing it to `object[key]` instead of `object.key` does not work also.
   ```yaml
   # does not work
   annotations:
      prometheus.scrape: <@ additionalInput.annotations.prometheus.scrape @>

   # also does not work
   annotations:
      prometheus.scrape: <@ additionalInput.annotations["prometheus.scrape"] @>
   ```

## Related issue(s)

Closes https://github.com/capactio/capact/issues/417

### DIFF
--- a/internal/cli/alpha/manifestgen/templates/helm-implementation.yaml.tmpl
+++ b/internal/cli/alpha/manifestgen/templates/helm-implementation.yaml.tmpl
@@ -1,0 +1,151 @@
+ocfVersion: 0.0.1
+revision: {{ .Revision }}
+kind: Implementation
+metadata:
+  prefix: "cap.implementation.{{ .Prefix }}"
+  name: {{ .Name }}
+  displayName: "{{ .Name }} Action"
+  description: "{{ .Name }} Action"
+  documentationURL: https://example.com
+  supportURL: https://example.com
+  maintainers:
+    - email: dev@example.com
+      name: Example Dev
+      url: https://example.com
+  license:
+    name: "Apache 2.0"
+
+spec:
+  appVersion: "1.0.x" # TODO(ContentDeveloper): Set the supported application version here
+  additionalInput:
+    parameters:
+      typeRef:
+        path: "cap.type.{{ .Prefix }}.{{ .Name }}-input"
+        revision: 0.1.0
+
+  outputTypeInstanceRelations:
+    config:
+      uses:
+        - terraform-release
+
+  implements:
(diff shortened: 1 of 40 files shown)


mszostok (commented):
Overall, really nice work 
good-job-clapping


It's awesome that you added test coverage and refactored code, it's more readable right now.

Regarding problem with escaping in jinja templating we can wait for our jinja sensei @lukaszo

Trojan295:
I reopened the PR in https://github.com/capactio/capact/pull/459, as I accidentally removed the branch from my fork...

pkosiec:
@Trojan295 FYI you could reopen the same pull request if you didn't create the new one 🙂  You can try closing the new one and reopening this, probably it would be easier to track post-review changes, but it's up to you.

internal/cli/alpha/manifestgen/templates/helm-implementation.yaml.tmpl
```diff
@@ -0,0 +1,154 @@
...
+  license:
+    name: "Apache 2.0"
+
+spec:
+  appVersion: "1.0.x" # TODO(ContentDeveloper): Set the supported application version here
+  additionalInput:
+    parameters:
+      typeRef:
+        path: "cap.type.{{ .Prefix }}.{{ .Name }}-input"
+        revision: 0.1.0
```
mszostok, line 24:
```suggestion
  additionalInput:
    parameters:
      additional-parameters:
        typeRef:
          path: "cap.type.{{ .Prefix }}.{{ .Name }}-input"
          revision: 0.1.0
```

after #450 syntax was changed.

PR #453, opened 2021-08-19
Replace gotest.tools with testify and unify commands naming

mszostok:
## Description

- Replace gotest.tools with testify
- Unify commands naming


mszostok:
will be merged as a part of #457

PR #454, opened 2021-08-19
Remove hardcoded images for Capact charts

mszostok:
## Description

Unfortunately, hardcoded images were not revert after merging to main. In a result, on all PRs, main, long-running cluster and local dev cluster has always the same engine image.

- Remove hardcoded images for Capact charts


pkosiec (approved):
tenor-106698694

PR #456, opened 2021-08-20
Explicitly set Minio access and secret key

Trojan295:
## Description

- Explicitly set Minio access and secret key

## Testing

Recreate cluster from this branch:
https://github.com/capactio/capact/actions/runs/1150357781

After that I run the latest, failed build from main and it updated the cluster successfully. Minio credentials aren't changed thanks to the reuse values.
https://github.com/capactio/capact/actions/runs/1150120521


internal/cli/capact/components.go
```diff
@@ -320,7 +321,31 @@ func (n *Neo4j) InstallUpgrade(ctx context.Context, version string) (*release.Re
...
+	minioValues["accessKey"] = accessKeyMap
+
+	secretKeyMap := make(map[string]interface{})
+	minioValues["secretKey"] = secretKeyMap
+
+	k8sClient, _ := a.configuration.KubernetesClientSet()
+	s, err := k8sClient.CoreV1().Secrets(a.opts.Namespace).Get(ctx, a.ReleaseName+"-minio", metav1.GetOptions{})
+
+	if apierrors.IsNotFound(err) {
+		// Chart does not exist, so we have to generate and set the credentials
```
pkosiec, line 338: This assumes the secret is always created along Minio installation. What if someone uses `existingSecret` with accessKey and secretKey? Maybe we should use it this way on our CI?

pkosiec (commented):
Let's discuss this offline, because I think we should go with a bit different approach - create secret separately and then reuse it for Minio installation. For upgrade, we should reuse the values (`existingSecret` is still provided).

internal/cli/capact/components.go
pkosiec: Okay, I checked the current behavior and we need to just pass access key and secret key during the installation if user hasn't provided existing secret.

The commands:
```
helm install my-release bitnami/minio --set accessKey.password=423432432423 --set secretKey.password=231213123123
helm upgrade my-release bitnami/minio
```
work flawlessly.


So, please refactor it to something like:

```go
func (a *Argo) InstallUpgrade(ctx context.Context, version string) (*release.Release, error) {
	upgradeCli := a.upgradeAction(version)


	values := a.Overrides
	// TODO: Split it to separate func
	if (existingSecret is not provided and this is very first Argo installation) { // TODO: Change to real condition instead of pseudocode 

		// Chart does not exist, so we have to generate and set the credentials

		var accessKey, secretKey string
		var err error

		
		accessKey, err = util.CryptoRandomAlphaNumeric(10)
		// TODO: handle error
		secretKey, err = util.CryptoRandomAlphaNumeric(40)
		// TODO: handle error


		values := map[string]interface{}{
			"minio": map[string]interface{}{
				"accessKey": map[string]interface{}{
					"password": accessKey,
				},
				"secretKey": map[string]interface{}{
					"password": secretKey,
				},
			},
		}


		values = tools.MergeMaps(values, a.Overrides)
	}
	
	return a.runUpgrade(upgradeCli, values)
}
```

pkosiec (commented):
Okay, as discussed offline, we don't support Minio overrides during upgrade, so we need to do it in CLI. But, it is all about first Argo installation - we need to pass accessKey and secretKey if user haven't provided `existingSecret` or their own accessKey/secretKey values.

internal/cli/capact/components.go
```diff
@@ -320,11 +321,93 @@ func (n *Neo4j) InstallUpgrade(ctx context.Context, version string) (*release.Re
 func (a *Argo) InstallUpgrade(ctx context.Context, version string) (*release.Release, error) {
 	upgradeCli := a.upgradeAction(version)
 
-	values := tools.MergeMaps(map[string]interface{}{}, a.Overrides)
+	values := make(map[string]interface{})
+
+	isReleased, err := a.isReleased()
```
pkosiec, line 326:
```suggestion
	installed, err := a.isInstalled()
```

internal/cli/capact/components.go
```diff
@@ -320,11 +321,93 @@ func (n *Neo4j) InstallUpgrade(ctx context.Context, version string) (*release.Re
...
+	if err != nil {
+		return nil, errors.Wrap(err, "while getting Argo Helm release")
+	}
+
+	if !isReleased && !a.areMinioCredentialsProvided(a.Overrides) {
```
pkosiec, line 331:
```suggestion
	if !installed && !a.areMinioCredentialsProvided(a.Overrides) {
```

pkosiec (approved):
Works well, just few suggestions to apply. Hope you'll like them 🙂

Issue #438, opened 2021-08-10
Rework Policy engine to support additionalTypeInstances and tidy up syntax


mszostok:
## Description

Currently, we are not able to pass additional TypeInstances in an obvious way. This task is about adjusting Policy to support more use-cases that are useful when working with Capact.

## Acceptance Criteria

There are three different areas to improve. We can treat them as subtasks:

- [x] Required TypeInstances (4SP) @pkosiec 
- [x] Additional TypeInstances (4SP) @pkosiec 
- [x] Additional Parameters (4SP) @mszostok

### Required TypeInstances (4SP)

1. Required input TypeInstances for Implementation (e.g. aws account): **3SP**
    - update documentation
    - consider if name should be specified or not
    - consider if TypeRef should be specified or not.  We can still have it but as a metadata, or just validate that the given typeInstance is of specified type. The reason for that is an example where user defines `implementationConstraints` base on attributes and doesn't know exact alias for a given TI. What he wants to do, is to simply say prefer AWS solutions and here is AWS account that should be used.
    ```yaml
    rules: # Configures the following behavior for Engine during rendering Action
      - interface: # Configures a second rule in the Action policy
          path: cap.interface.database.mongodb.install
        oneOf:
          - implementationConstraints:
              path: "cap.implementation.bitnami.mongodb.install"
            inject:
              # additionalParameters - for `additionalInput.parameters` 
              # additionalTypeInstances: - for `additionalInput.typeInstances`
              requiredTypeInstances: - for `requires`
                  - id: uuid
                     description: "GCP SA" # optional
    ```
1. Remove `apiVersion` from Policy
1.  Update `ListImplementationRevisionForInterface` to ignore Implementation which has  `requiredTypeInstances` but user didn't specify them in Policy. **1SP**, split to: https://github.com/capactio/capact/issues/358

### Additional TypeInstances (4SP)

1. Input TypeInstances specified for the Interface are used only for Interface Implementation. 
   -  Remember to update the validation process, as currently we ignore unknown TypeInstances. **1SP**
1. Additional Input TypeInstances for Implementation (e.g. already existing database) are specified under policy. **3SP**
    - update documentation
    ```yaml
      - interface: # Configures a second rule in the Action policy
          path: cap.interface.database.mongodb.install
        oneOf:
          - implementationConstraints:
              path: "cap.implementation.bitnami.mongodb.install"
            inject:
              additionalTypeInstances:
                 - name:  "name"
                    id: ${TI_ID}
    ```

[...]

PR #458, opened 2021-08-23
Filter Implementations which need missing required TypeInstance injection

pkosiec:
## Description

- Filter Implementations which need missing required Type Instance injection
- Refactor Resolving Type References for Policy

## Notes

~This pull request bases on #452. Once #452 is merged, this PR will be rebased.~
Done

## Testing

No manual testing needed, as modified integration tests and unit tests (including renderer one) cover this case.

## Related issue(s)

https://github.com/capactio/capact/issues/438

### DIFF
--- a/pkg/engine/k8s/policy/deepcopy.extend.go
+++ b/pkg/engine/k8s/policy/deepcopy.extend.go
@@ -1,0 +1,52 @@
+package policy
+
+import "capact.io/capact/internal/maps"
+
+// DeepCopyInto writes a deep copy of AdditionalParametersToInject into out.
+// controller-gen doesn't support interface{} so writing it manually
+func (in *AdditionalParametersToInject) DeepCopyInto(out *AdditionalParametersToInject) {
+	*out = *in
+	out.Value = maps.Merge(out.Value, in.Value)
+}
+
+// DeepCopy returns a new deep copy of AdditionalParametersToInject.
+// controller-gen doesn't support interface{} so writing it manually
+func (in *AdditionalParametersToInject) DeepCopy() *AdditionalParametersToInject {
+	if in == nil {
+		return nil
+	}
+	out := new(AdditionalParametersToInject)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopy returns a new deep copy of InjectData.
+// controller-gen doesn't support interface{} so writing it manually
+func (in *InjectData) DeepCopy() *InjectData {
+	if in == nil {
+		return nil
+	}
+	out := new(InjectData)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyInto writes a deep copy of InjectData into out.
+// controller-gen doesn't support interface{} so writing it manually
+func (in *InjectData) DeepCopyInto(out *InjectData) {
+	*out = *in
+	if in.RequiredTypeInstances != nil {
+		in, out := &in.RequiredTypeInstances, &out.RequiredTypeInstances
+		*out = make([]RequiredTypeInstanceToInject, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+	if in.AdditionalParameters != nil {
+		in, out := &in.AdditionalParameters, &out.AdditionalParameters
+		*out = make([]AdditionalParametersToInject, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+}
--- a/pkg/engine/k8s/policy/resolve_test.go
+++ b/pkg/engine/k8s/policy/resolve_test.go
@@ -1,0 +1,65 @@
+package policy_test
+
+import (
+	"context"
+	"testing"
+
+	"capact.io/capact/internal/ptr"
+	"capact.io/capact/pkg/engine/k8s/policy"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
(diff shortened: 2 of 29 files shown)


pkg/engine/k8s/policy/deepcopy.extend.go
```diff
@@ -0,0 +1,52 @@
+package policy
```
pkosiec, line 1: To discuss the file name: I wanted to set `zz_generated.deepcopy.extend.go`, but it may be confusing that it's generated. No strong opinion. WDYT?

pkg/hub/client/public/filter_test.go
```diff
@@ -159,9 +160,127 @@ func TestImplementationRequirementsSatisfiedByFilters(t *testing.T) {
...
-					TypeRef: &gqlpublicapi.TypeReferenceWithOptionalRevision{
+					TypeRef: &gqlpublicapi.TypeReferenceInput{
 						Path:     "cap.type.gcp.sa",
-						Revision: ptr.String("0.1.1"),
+						Revision: "0.1.1",
+					},
+				},
+			},
+		},
+		{
```
mszostok, line 170: What about such scenario:
```go
{
			name: "Return Implementations satisfied by injected RequiredTypeInstances",
			expRevision: []gqlpublicapi.ImplementationRevision{
				fixImplementationRevision("without-any-requirements", "0.0.1"),
				fixImplementationRevisionWithRequire("with-gcp-sa-requirement", "0.1.0", gqlpublicapi.ImplementationRequirement{
					AllOf: []*gqlpublicapi.ImplementationRequirementItem{
						{
							Alias: ptr.String("gcp-sa"),
							TypeRef: &gqlpublicapi.TypeReference{
								Path:     "cap.type.gcp.sa",
								Revision: "0.1.1",
							},
						},
						{
							TypeRef: &gqlpublicapi.TypeReference{
								Path:     "cap.core.type.platform.kubernetes",
								Revision: "0.1.1",
							},
						},
					},
				}),
			},
			requiredTIInjectionSatisfiedBy: []*gqlpublicapi.TypeInstanceValue{
				{
					TypeRef: &gqlpublicapi.TypeReferenceInput{
						Path:     "cap.type.gcp.sa",
						Revision: "0.1.1",
					},
				},
			},
			requirementsSatisfiedBy: []*gqlpublicapi.TypeInstanceValue{
				{
					TypeRef: &gqlpublicapi.TypeReferenceInput{
						Path:     "cap.core.type.platform.kubernetes",
						Revision: "0.1.1",
					},
				},
			},
		},
```

I understand that it currently it's additional level but maybe it should be also complementary. So if we see sth like that 
- we simple add `requiredTIInjectionSatisfiedBy` into `requirementsSatisfiedBy` to pass our check, 
- Or change the implementation but this will change also API: (not 100% sure about that approach)
```go
for _, item := range implReq {
		if item.TypeRef == nil {
			continue
		}

		// required TypeInstance injection
		if item.Alias != nil {
			if !contains(requiredTIInjectionSatisfiedBy, *item.TypeRef) {
				// injection needed and TypeInstance for injection not found
				return false
			}
		} else {
			if !contains(availableReq, *item.TypeRef) {
				// all needs to be satisfied, so we can already give up
				return false
			}
		}
	}
```

pkg/hub/client/public/filter_test.go
```diff
@@ -159,9 +160,127 @@ func TestImplementationRequirementsSatisfiedByFilters(t *testing.T) {
...
+			name: "Return Implementations satisfied by injected RequiredTypeInstances",
+			expRevision: []gqlpublicapi.ImplementationRevision{
+				fixImplementationRevision("without-any-requirements", "0.0.1"),
+				fixImplementationRevisionWithRequire("with-gcp-sa-requirement", "0.1.0", gqlpublicapi.ImplementationRequirement{
+					Prefix: "cap.core.type.platform",
```
mszostok, line 175: This prefix doesn't fit here. I checked and it's unused. Maybe it will be better to get rid of it because it was really confusing for me as based on the API I tried to create such fixture:
```go
				Prefix: "cap.core.type.platform",
					AllOf: []*gqlpublicapi.ImplementationRequirementItem{
						{
							TypeRef: &gqlpublicapi.TypeReference{
								Path:     "kubernetes",
								Revision: "0.1.1",
							},
						},
					},
```

mszostok (approved):

mszostok:
I approved that as it works and it lgtm. I'm only not sure about the API and it's behaviour but we can discuss that offline.

pkg/engine/k8s/policy/resolve_test.go
```diff
@@ -0,0 +1,65 @@
...
+			Input:    fixComplexPolicyWithTypeRef(),
+			Expected: fixComplexPolicyWithTypeRef(),
+		},
+	}
+
+	for _, testCase := range tests {
+		tc := testCase
+		t.Run(tc.Name, func(t *testing.T) {
+			// when
+			in := tc.Input
```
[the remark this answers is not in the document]
pkosiec: It's not 😄  removing

pkg/hub/client/public/filter_test.go
pkosiec: As discussed offline, I will merge all provided `requiredTIInjectionSatisfiedBy` into `requirementsSatisfiedBy` to cover the following use cases:
- the one listed above 
- using `requiredTIInjectionSatisfiedBy` without `requirementsSatisfiedBy`
Also, I will add unit tests for these two cases.

mszostok (approved):

PR #460, opened 2021-08-25
Fix k8s controller test, add retry for createObject, revert hub-manifest sources

mszostok:
## Description

- Fix k8s controller test,
- Add retry for createObject, 
- Revert hub-manifest sources

## Related issue(s)

- https://github.com/capactio/capact/issues/438

### DIFF
--- a/internal/cli/capact/components.go
+++ b/internal/cli/capact/components.go
@@ -16,6 +16,7 @@ import (
 	"capact.io/capact/internal/ptr"
 
 	util "github.com/Masterminds/goutils"
+	"github.com/avast/retry-go"
 	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart/loader"
@@ -28,11 +29,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/strings/slices"
 )
 
-// Component is a Capact component which can be installed in the environement
+// Component is a Capact component which can be installed in the environment
 type Component interface {
 	InstallUpgrade(ctx context.Context, version string) (*release.Release, error)
 	Name() string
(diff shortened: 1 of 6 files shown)


internal/cli/capact/components.go
```diff
@@ -648,12 +645,23 @@ func createObject(configuration *action.Configuration, content []byte) error {
...
-			return nil
-		}
-		return errors.Wrapf(err, "while creating the object")
+	// 5 exponential retries: ~2µs ~12ms ~70ms ~325ms ~1.63s
+	retryBackoff := retry.DefaultBackoff
+	retryBackoff.Steps = 5
+	err = retry.OnError(retryBackoff, differentThanAlreadyExistErr, func() error {
+		_, err = configuration.KubeClient.Create(res)
+		// You have to return err itself here (not wrapped inside another error)
+		// so we can identify it correctly.
```
pkosiec, line 654:
```suggestion
		// Return not wrapped error itself here so we can identify it correctly
```

internal/cli/capact/components.go
```diff
@@ -648,12 +645,23 @@ func createObject(configuration *action.Configuration, content []byte) error {
...
+		return err
+	})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
```
pkosiec, line 657: If we have such function already... 😄 

```suggestion
	if err != nil && differentThanAlreadyExistErr(err error) {
```

internal/k8s-engine/controller/suite_test.go
```diff
@@ -1,4 +1,5 @@
-//+build controllertests
+//go:build controllertests
+// +build controllertests
```
mszostok, line 2: should I revert it? It's added because of go 1.17 which I use locally + in my IDE (which added that automatically) 

>The go command now understands //go:build lines and prefers them over // +build lines. The new syntax uses boolean expressions, just like Go, and should be less error-prone. As of this release, the new syntax is fully supported, and all Go files should be updated to have both forms with the same meaning. To aid in migration, gofmt now automatically synchronizes the two forms. For more details on the syntax and migration plan, see https://golang.org/design/draft-gobuild.



source: https://golang.org/doc/go1.17

internal/cli/capact/components.go
```diff
@@ -648,12 +645,23 @@ func createObject(configuration *action.Configuration, content []byte) error {
...
-	if _, err := configuration.KubeClient.Create(res); err != nil {
-		// If the error is CRD already exists, return.
-		if apierrors.IsAlreadyExists(err) {
-			return nil
-		}
-		return errors.Wrapf(err, "while creating the object")
+	// 5 exponential retries: ~2µs ~12ms ~70ms ~325ms ~1.63s
+	retryBackoff := retry.DefaultBackoff
+	retryBackoff.Steps = 5
+	err = retry.OnError(retryBackoff, differentThanAlreadyExistErr, func() error {
```
pkosiec, line 651: 🤔 Usually we use `github.com/avast/retry-go` in such scenario. There are two places where we use `k8s.io/client-go/util/retry` - here and in `k8s.go` file, also for installation.

We can keep it as it is, but I think long-term it would be great to have consistency and use just a single package for a single thing. Same what we have with multierrors.

internal/cli/capact/components_test.go
```diff
@@ -0,0 +1,124 @@
...
+	objToCreate := fmt.Sprintf(issuerTemplate, clusterIssuerName, certManagerSecretName)
+
+	// when
+	startedCall := time.Now()
+	err := createObject(actionCfg, []byte(objToCreate))
+
+	// then
+	assert.EqualError(t, err, internalError.Error())
+	assert.Equal(t, 5, failingCli.CreateCallsCnt)
+	assert.InDelta(t, expTotalRetryDurationInSec, time.Since(startedCall).Seconds(), 0.2)
```
pkosiec, line 79: Not sure if we should care that much about the time duration. Personally, I would remove this assertion. Another thing is that even if we don't do actual calls, I think there might be some rare VM slowdowns which could potentially cause this test to fail.

internal/k8s-engine/controller/suite_test.go
pkosiec: We can keep it or remove it. We need to migrate to Go 1.17 in the near future both locally and on CI anyway, so we can add the `//go:build` lines as a part of that small bump.

But, if you decide to keep it, then please also update all other files where we use build tags 👍

pkosiec (approved):
LGTM

Issue #353, opened 2021-06-16
Add verbose flag to capact CLI


platten:
## Description
Add `--verbose` or `-v` flag to command line to print out what is happening under the hood when running commands in the capact CLI.

## AC 
- Add global verbose flag to CLI which can be used by all components without redefining,
- Add verbose mode for GraphQL client,
- Add spent time info to each finished step (if verbose mode enabled).

Verbose should work in two modes:
- `-v` same as `-v=1` (default) it logs only "verbose" data (human-friendly)
- `-v=2` it logs additional trace data, e.g. request headers, body, response header, body.

## Reason

- Aid in diagnosing any potential issues with Capact and provide additional information for filing more useful bug reports.
- See GraphQL queries as they are being sent out and the raw responses received. This may help with understanding how the CLI operates and can help with developing additional clients and integrations.

PR #462, opened 2021-08-26
Add global verbose flag to CLI, add spent time info

mszostok:
## Description

- Add global verbose flag to CLI, 
- Add verbose mode for GraphQL client,
- Add spent time info to each finished step (if verbose mode enabled),
- Add unit-test coverage for flag.

It is a small change, see first commit: https://github.com/mszostok/capact/commit/456f774ebf76dd3cad740a40e0d2e52a29a21500

**rest is just updated docs (autogenerated)**

**Screenshot:** _(click to fullscreen)_

<p align="center">
  logo
</p>

**Recording:** https://asciinema.org/a/432543
### REASON:
- sometimes we need to debug why CLI failed. Last time we had such issue: 
  ```
  ❯ capact action create \
    cap.interface.database.postgresql.install \
    --name rds \
    --parameters-from-file ./hidden/manifest-values/rds/params.yaml
  
  Error: 1 validation error detected:
  - Parameters "input-parameters":
      * Unknown parameter. Cannot validate it against JSONSchema.
  
  
  ❯ cat ./hidden/manifest-values/rds/params.yaml
  superuser:
    username: foo
    password: bar
  ```
  
  and it was wired that it failed, but with enabled GraphQL client logs it was easy to find out the root cause
  ```bash
  {"data":{"interface":{"rev":{"spec":{"input":{"parameters":[{"jsonSchema":null,"name":"input-parameters"}],"typeInstances":[]}}}}}}
  ```
  as you can see in body we didn't have `typeRef`

  but I had to manually add a logger to the client and build our Capact binary. Much easier it will be to tell someone to just add `-v` flag and share the response.

- last time I was debugging our installation time, and I was not able to tell how long it took, and when was started. It's nice to know if sth was stuck and also know approximately the time for each step (which you can get from previous runs). 

### IDEA:

Currently, we had verbose only for two commands (validation, install). We used the local variable assigned to a given `Options`. It's good, but it complicates a lot as we need to pass it down to each layer.

I decided to add a `VerboseMode` variable for `cli` pkg. Thanks to that, `cli` packages can use it in directly, as it is populated on `root` command.

The rule is that it can be used directly only under `cli` pkg. If you call external functions from e.g. `pkg/` you should pass this value and don't refer from e.g. `pkg` to `cli.VerboseMode` as it's tightly coupled with CLI and `flags` system.

## Related issue(s)

- https://github.com/capactio/capact/issues/353

### DIFF
--- a/internal/cli/verbose_test.go
+++ b/internal/cli/verbose_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestVerboseModeFlag(t *testing.T) {
+func TestVerboseModeFlagHappPath(t *testing.T) {
 	tests := map[string]struct {
 		givenRawOption string
 		expectedMode   cli.VerboseModeFlag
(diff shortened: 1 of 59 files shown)


mszostok:
As discussed f2f with @pkosiec I had to change the `verbose` option. Previously it was just a `bool` but  after adding additional data to verbose mode it broke the validate command
```
# without verbose mode it prints only errors
capact manifest validate --server-side ../hub-manifests/**/*.yaml
# with verbose mode it prints all checked files
capact manifest validate --server-side -v ../hub-manifests/**/*.yaml
```

If we will stay with `bool` then we will print also http calls in `capact manifest validate --server-side -v ../hub-manifests/**/*.yaml` making it unreadable in normal scenario. As a result, we introduced the verbose levels. 

We know that the verbose levels from k8s (`-v=1-10`) are problematic for others as it's hard to remember e.g. the difference between `5` and `8`. For us we decided to simplify that, to have only two modes/levels:
- `-v` same as `-v=1` (default) it logs only "verbose" data (it is still human-friendly but prints more info)
- `-v=2` it logs additional trace data, e.g. request headers, body, response header, body.

so to get http calls for validation you need to run:
```
# with verbose mode it prints all checked files and executed calls
capact manifest validate --server-side -v=2 ../hub-manifests/**/*.yaml
```

mszostok:
allowed options:

```
simple: "-v", "-v=1", "-v=simple"
tracing: "-v=2", "-v=tracing"
disabled: "-v=disable" or just no flag
```

internal/cli/verbose_test.go
```diff
@@ -0,0 +1,68 @@
...
+			cli.RegisterVerboseModeFlag(flags)
+
+			var args []string
+			if tc.givenRawOption != "" {
+				args = append(args, tc.givenRawOption)
+			}
+
+			// when
+			err := flags.Parse(args)
+			require.NoError(t, err)
```
lukaszo, line 62: Maybe let's also test wrong values. `-v=3` What do we expect in such case?

Issue #468, opened 2021-08-30
Improve handling input parameters by Action CR


mszostok:
## Description

Fix handling input parameters by Action CR. Currently, we can specify only one input parameters and not collection. 

Interface:
https://github.com/capactio/capact/blob/0fbf562b0a3cc452a43f597ff17b55236a3d53e2/ocf-spec/0.0.1/examples/interface.yaml#L20-L24

Engine/Action:

https://github.com/capactio/capact/blob/0fbf562b0a3cc452a43f597ff17b55236a3d53e2/pkg/engine/k8s/api/v1alpha1/action_types.go#L152-L154

https://github.com/capactio/capact/blob/0fbf562b0a3cc452a43f597ff17b55236a3d53e2/pkg/sdk/renderer/argo/helpers.go#L165-L188

https://github.com/capactio/capact/blob/0fbf562b0a3cc452a43f597ff17b55236a3d53e2/pkg/sdk/renderer/argo/renderer.go#L18-L19

Capact CLI should accept parameters collection. Think also about backward compatibility. 

AC:
- Support multiple input parameters across all our API
- Update documentation (also, fix **all** tutorials)

## Reason

Our API is not consistent and is not readable. Additionally, we narrow-down usage of already working feature.

```bash
❯ capact action create \
  cap.interface.database.postgresql.install \
  --name rds \
  --parameters-from-file ./hidden/manifest-values/rds/params.yaml

Error: 1 validation error detected:
- Parameters "input-parameters":
    * Unknown parameter. Cannot validate it against JSONSchema.


❯ cat ./hidden/manifest-values/rds/params.yaml
superuser:
  username: foo
  password: bar
```

As you can see, user gets info about `input-parameters` which is hardcoded name in engine. It's hard for end-user to understand the flow.

PR #470, opened 2021-08-31
Support multiple input parameters

Trojan295:
## Description

- Allow providing multiple user input parameters through CLI
- Inject multiple input parameters in rendering

## Testing

AFAIK we don't have any Interface with two input parameters. I added a test for this + you can run an existing Action, eg. setup a Bitnami Helm Postgresql using the following input parameters:

```yaml
input-parameters:
  superuser:
    username: damian
    password: sup3rs3cr3t
  defaultDBName: postgres
```

Check also `capact upgrade` on a cluster from this PR and CLI compiled from this PR:
```bash
capact upgrade --action-name-prefix 'capact-upgrade-' --version @latest --helm-repo-url @latest --increase-resource-limits --wait
```

## Related issue(s)

Resolves https://github.com/capactio/capact/issues/468

### DIFF
--- a/internal/cli/upgrade/upgrade.go
+++ b/internal/cli/upgrade/upgrade.go
@@ -28,6 +28,8 @@ const (
 	capactTypeRefPath          = "cap.type.capactio.capact.config"
 	helmReleaseTypeRefPath     = "cap.type.helm.chart.release"
 
+	actionParametersKey = "input-parameters"
+
 	randomSuffixLength = 5
 	pollInterval       = time.Second
 )
@@ -375,7 +377,7 @@ func mapToInputTypeInstances(capactCfg gqllocalapi.TypeInstance) ([]*gqlengine.I
 
 func mapToInputParameters(params capact.InputParameters) (gqlengine.JSON, error) {
 	parameters := map[string]interface{}{
-		"input-parameters": params,
+		actionParametersKey: params,
 	}
 
 	marshalled, err := json.Marshal(parameters)
--- a/internal/cli/upgrade/upgrade.go
+++ b/internal/cli/upgrade/upgrade.go
@@ -374,7 +374,11 @@ func mapToInputTypeInstances(capactCfg gqllocalapi.TypeInstance) ([]*gqlengine.I
 }
 
 func mapToInputParameters(params capact.InputParameters) (gqlengine.JSON, error) {
-	marshalled, err := json.Marshal(params)
+	parameters := map[string]interface{}{
+		"input-parameters": params,
+	}
+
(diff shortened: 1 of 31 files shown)


internal/cli/upgrade/upgrade.go
```diff
@@ -374,7 +374,11 @@ func mapToInputTypeInstances(capactCfg gqllocalapi.TypeInstance) ([]*gqlengine.I
 }
 
 func mapToInputParameters(params capact.InputParameters) (gqlengine.JSON, error) {
-	marshalled, err := json.Marshal(params)
+	parameters := map[string]interface{}{
+		"input-parameters": params,
```
lukaszo, line 378: Maybe make it a constant?

pkg/sdk/renderer/argo/dedicated_renderer.go
```diff
@@ -816,6 +775,85 @@ func (r *dedicatedRenderer) sleepContainer() *apiv1.Container {
...
+
+func (r *dedicatedRenderer) jqContainer(inputFilepath, jqPath, outputFilepath string) *apiv1.Container {
+	return &apiv1.Container{
+		Image: "ghcr.io/capactio/infra/jq:1.6",
+		// Built using:
+		// FROM alpine:3.14.2
+		// RUN apk add jq=1.6-r1
+
+		Command: []string{"sh", "-c"},
+		Args:    []string{fmt.Sprintf(`cat "%s" | jq -r '%s' | tee "%s" && sleep 1`, inputFilepath, jqPath, outputFilepath)},
```
lukaszo, line 853: Maybe instead of using jq to extract parameters, we can just store then in secrets as separate keys?
Instead of:
```yaml
apiVersion: v1
data:
  parameters.json: eyJleHRyYS1wYXJhbWV0ZXJzIjp7ImRlZmF1bHREQk5hbWUiOiJleHRyYS1kYmRiIiwic3VwZXJ1c2VyIjp7InBhc3N3b3JkIjoiZXh0cmEtcGFzczEyMyIsInVzZXJuYW1lIjoiZXh0cmEifX0sImlucHV0LXBhcmFtZXRlcnMiOnsiZGVmYXVsdERCTmFtZSI6ImRiZGIiLCJzdXBlcnVzZXIiOnsicGFzc3dvcmQiOiJjYXBhY3QtcGFzczEyMyIsInVzZXJuYW1lIjoiY2FwYWN0In19fQ==
```

We could do:
```yaml
apiVersion: v1
data:
  input-parameters: <base64>
  another-parameters: <base64>
```

lukaszo (approved):

Issue #472, opened 2021-09-02
`capact act watch` cannot be stopped


lukaszo:
## Description

`capact act watch` cannot be stopped using ctrl-c

## Expected behavior

should stop

## Actual behavior

doesn't stop :)

## Steps to reproduce

* create any action
* run it
* watch it
* try to stop it

pkosiec:
We use argo watch command directly and looks like they don't use `Context` from Cobra command:

```go
		Run: func(cmd *cobra.Command, args []string) {
			if len(args) != 1 {
				cmd.HelpFunc()(cmd, args)
				os.Exit(1)
			}
			ctx, apiClient := client.NewAPIClient() // <-- Here
			serviceClient := apiClient.NewWorkflowServiceClient()
			namespace := client.Namespace()
			watchWorkflow(ctx, serviceClient, namespace, args[0], getArgs)
		}, 
```

Going few levels deeper, we have:

```go
	gatekeeper, err := auth.NewGatekeeper(auth.Modes{auth.Server: true}, clients, restConfig, nil, auth.DefaultClientForAuthorization, "unused")
	if err != nil {
		return nil, nil, err
	}
	ctx, err := gatekeeper.Context(context.Background())
```

And what gatekeeper does is adding few values:

```go
func (s *gatekeeper) Context(ctx context.Context) (context.Context, error) {
	clients, claims, err := s.getClients(ctx)
	if err != nil {
		return nil, err
	}
	ctx = context.WithValue(ctx, WfKey, clients.Workflow)
	ctx = context.WithValue(ctx, EventSourceKey, clients.EventSource)
	ctx = context.WithValue(ctx, SensorKey, clients.Sensor)
	ctx = context.WithValue(ctx, KubeKey, clients.Kubernetes)
	ctx = context.WithValue(ctx, ClaimsKey, claims)
	return ctx, nil
}
```

So we can't just replace the context with the `cmd.Context()`, we would need to merge it. As the [`context.Merge` proposal is still open](https://github.com/golang/go/issues/36503), we would need to use something like https://github.com/teivah/onecontext.

Or, figure some other way 🙂

Trojan295:
I created a PR in Argo, which would fix this problem:
https://github.com/argoproj/argo-workflows/pull/6667

PR #473, opened 2021-09-02
Fix context cancellation in capact action watch

Trojan295:
## Description

- Fix problem with canceling Argo watch

I also created a PR in Argo, which would resolve this problem:
https://github.com/argoproj/argo-workflows/pull/6667

## Related issue(s)

Resolves https://github.com/capactio/capact/issues/472


lukaszo:
Now I see some red text. It wasn't there before. image

Can we get rid of it?

Also, can we get rid of the warning about "Security context"? I know it's not a part of the task, but if it's easy do remove then maybe let's try.

lukaszo (approved):

PR #474, opened 2021-09-03
Show sudo password prompt

### LINKED ISSUE #412: It's not clear that user needs to provide a sudo password when updating /etc/hosts file
## Description

When `capact install` command is used, it's not clear that the user needs to provide a sudo password. The prompt may be hidden, on Mac systems only a lock icon is seen.

We need a clear way to ask a user for a password.

Solutions to consider:

1. Move updating /etc/hosts file and certificates to a new command which needs to be run by privileged user.
2. Make it clear that sudo will be needed, by writing, for example: "Entering sudo mode" 
3. Asking always for a password and passing it to the sudo command.

lukaszo:
## Description

- Remove hiding sudo promt
- Disable spinner when sudo may be required

## Testing

1. Remove `gateway.capact.local` entry from `/etc/hosts`
2. Build CLI `make build-tool-cli`
3. Run `capact` command without installing any components `capact install --install-component=""                                                                                                                             `

## Related issue(s)

Closes #412

### DIFF
--- a/internal/cli/capact/os.go
+++ b/internal/cli/capact/os.go
@@ -24,7 +24,7 @@ func AddGatewayToHostsFile() error {
 		return nil
 	}
 
-	fmt.Printf("   * Updating %s file. Entering sudo password may be required \n", hosts)
+	fmt.Printf("   * Updating %s file. Entering sudo password may be required\n", hosts)
 	// #nosec G204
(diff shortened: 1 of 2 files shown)


mszostok (approved):
Tested 3 scenarios:
- empty /etc/hosts
- no certs
- no certs + empty /etc/hosts 

worked properly in all of them 👍

Issue #423, opened 2021-08-03
Retry mechanism in Engine controller loop doesn't work 


mszostok:
## Description

The retry mechanism in Engine controller loop doesn't work properly, as you can see in logs:
```
  2021-08-03T12:07:08.474+0200	DEBUG	controller-runtime.manager.events	zapr@v0.1.0/zapr.go:69	Warning	{"object": {"kind":"Action","namespace":"default","name":"e2e-test-1-8081","uid":"0cc0a040-4fe8-445a-a054-da9ab46832a5","apiVersion":"core.capact.io/v1alpha1","resourceVersion":"5761139"}, "reason": "ReadyToRun", "message": "Cannot create runner ServiceAccount: while getting ClusterRoleBinding for update: ClusterRoleBinding.rbac.authorization.k8s.io \"e2e-test-1-8081\" not found (will retry - 1/15)"}
  2021-08-03T12:07:08.618+0200	INFO	controllers.Action	zapr@v0.1.0/zapr.go:69	Execute runner	{"action": "default/e2e-test-1-8081"}
  2021-08-03T12:07:08.752+0200	DEBUG	controller-runtime.manager.events	zapr@v0.1.0/zapr.go:69	Warning	{"object": {"kind":"Action","namespace":"default","name":"e2e-test-1-8081","uid":"0cc0a040-4fe8-445a-a054-da9ab46832a5","apiVersion":"core.capact.io/v1alpha1","resourceVersion":"5761142"}, "reason": "ReadyToRun", "message": "Cannot create runner ServiceAccount: while getting ClusterRoleBinding for update: ClusterRoleBinding.rbac.authorization.k8s.io \"e2e-test-1-8081\" not found (will retry - 2/15)"}
  2021-08-03T12:07:08.839+0200	INFO	controllers.Action	zapr@v0.1.0/zapr.go:69	Execute runner	{"action": "default/e2e-test-1-8081"}
[...]

PR #477, opened 2021-09-06
Filter out Action Status update event

lukaszo:
## Description

- Filter out Action Status update event
  * Add `Result{Requeue: True}` when Request shoul be requeued with a backoff
  * Add `Resukt{RequeueAftert: 1*time.Microsecond}` when Request should be
  requeued without backoff

## Testing

* Run the cluster with populator disabled:

  ```bash
  DISABLE_MONITORING_INSTALLATION=true ENABLE_POPULATOR=false make dev-cluster
  ```

* Make a typo in the bitnami Postgresql manifest. For example, change `{{steps` to `{{stepssss`.
* Populate manifests
* Create and action: 

  ```bash
  capact action create   cap.interface.database.postgresql.install   --name psql
  ```
  imput params:
  ```yaml
  superuser:
    username: capact
    password: capact-pass123
  defaultDBName: dbdb
   ```
* get the action status
  ```bash
  capact act get psql -o yaml
  ```
* monitor how fast retries are changing. The last few should take noticeable amount of seconds.

## Related issue(s)

Fixes: #423


mszostok (approved):
lgtm 👍

Issue #466, opened 2021-08-30
Capact on k3d with local registry and pre-loaded images


mszostok:
## Description

Add option to install Capact on k3d. 

**AC:**

- Add option to create k3d environment `capact env create k3d`
- Add to use it in our local `make dev-cluster` scripts
- Add option to "snapshot" docker images from Capact Helm Charts (create .tar file)
   -  support `@latest, version, local`
- Add option to use local registry with k3d
  -  this speeds up our local development
- Reduce neo4j startup time
- Reduce e2e test image size
- Documentation

## Reason

`kind` is good for testing Capact on pipelines. K3d is also a well known alternative. It's also a bit faster and allows easily to mount already existing images, which means you don't need to download them each time from the internet.

Additionally, local registry simplify pushing and speeds up local development.

PR #480, opened 2021-09-10
Make `capact install` "internetless" if needed

mszostok:
## Description

- Add opt to use local Helm charts. Using `--helm-repo` you can specify relative path to current working directory or URL.

  ```bash
  # Install with locally pulled Helm charts, version is required
  capact install --version 0.4.0 --helm-repo charts/capact
  
  # Install with local Helm charts
  capact install --version @local
  ```
- Change `tracing` to `trace` to unify naming, `--verbose=simple/trace/disable`
- Change `--helm-repo-url` to `--helm-repo` as now it can be path or URL in case of install. As the upgrade is executed by Capact Action it still needs to download charts from the internet.
- Print msg `Trusting self-signed CA certificate (..)` only if needed.
- Change ingress folder name to match the chart name
- Update populator docker images to contain local hub-manifests fetched from git. 

## Testing

Described here: https://github.com/capactio/website/pull/61

## Related issue(s)

Resolves https://github.com/capactio/capact/issues/466


internal/cli/environment/create/k3d_defaults.go
```diff
@@ -9,7 +9,7 @@ var k3dDefaultConfigTmpl string
 
 const (
 	// k3dDefaultNodeImage defines default Kubernetes image for a new k3d cluster.
-	k3dDefaultNodeImage = "docker.io/rancher/k3s:v1.20.7-k3s1"
+	k3dDefaultNodeImage = "docker.io/rancher/k3s:v1.20.10-k3s1"
```
lukaszo, line 12: Can you update it as well for kind? So, we use the same version in both cases? Kind defaults to 1.19.

Perfectly, we could have a base version in a one variable.

mszostok: I updated kind to 1.20.7 because they don't have 1.20.10, so I didn't introduce one variable for it.

lukaszo (approved):

PR #515, opened 2021-09-21
Send Slack notifications only on failures

lukaszo:
## Description

- Send Slack notifications only on failures
  This will ignore canceled, skipped and success results.

Possible results: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#needs-context

### DIFF
--- a/.github/workflows/branch-build.yaml
+++ b/.github/workflows/branch-build.yaml
@@ -279,11 +279,11 @@ jobs:
   slackNotification:
     name: Slack Notification
     runs-on: ubuntu-latest
-    needs: [ update-cluster ]
+    needs: [ update-cluster, integration-tests ]
     if: always()
     steps:
       - name: Slack Notification
-        if: ${{ always() && needs.update-cluster.result != 'success' }}
+        if: failure()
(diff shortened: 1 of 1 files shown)


.github/workflows/branch-build.yaml
```diff
@@ -279,11 +279,11 @@ jobs:
   slackNotification:
     name: Slack Notification
     runs-on: ubuntu-latest
-    needs: [ update-cluster ]
+    needs: [ update-cluster, integration-tests ]
     if: always()
     steps:
       - name: Slack Notification
-        if: ${{ always() && needs.update-cluster.result != 'success' }}
+        if: ${{ needs.update-cluster.result == 'failure' || needs.integration-tests.result == 'failure' }}
```
pkosiec, line 286: 👍 I think it makes sense to post Slack updates also on failed integration tests on main branch. 
But maybe we can use simply `${{ failure() }}` here?

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#failure

```suggestion
        if: ${{ failure() }}
```

pkosiec (approved):

PR #525, opened 2021-09-30
Update tests, enable tests after changing template

lukaszo:
## Description

- Update tests, enable tests after changing template


lukaszo:
Self merging as need to fix CI

PR #549, opened 2021-11-02
Fix `hack/eks/install.sh` and add server auth mod for Argo

mszostok:
## Description

- Fix ingress controller Helm chart path
- Gateway Ingress was manged by public ingress controller, where the entry in Route 53 is hard-coded to the internal ELB. The reason was the lack of `gateway.ingress.annotations.class=capact` override during install.
- Add missing region to aws CLI
- Bump EKS version to be compatible with Capact. Currently, you will get such error:
    ```bash
     ✗ Installing capact Helm chart
    Error: All attempts fail:
    #1: while installing Helm chart [capact]: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"
    #2: while installing Helm chart [capact]: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"
    #3: while installing Helm chart [capact]: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"
    ```
- The Argo server used to start with auth mode of "server" by default, but since v3.0 it defaults to the "client". As a result, you need to always pass the token (first generate it via SA). I changed it back to `server` as we don't expose Argo publically.

### DIFF
--- a/hack/eks/install.sh
+++ b/hack/eks/install.sh
@@ -77,7 +77,7 @@ capact::aws::install::capact() {
   shout "Deploying Capact..."
   capact install --environment eks \
     --version "${CAPACT_VERSION}" \
-    --capact-overrides "global.domainName=${CAPACT_DOMAIN_NAME}" \
+    --capact-overrides "global.domainName=${CAPACT_DOMAIN_NAME},gateway.ingress.annotations.class=capact" \
     --cert-manager-overrides "${CUSTOM_CERT_MANAGER_OVERRIDES}"
 
   shout "Capact deployed successfully!"
--- a/hack/eks/install.sh
+++ b/hack/eks/install.sh
@@ -127,12 +127,12 @@ capact::aws::register_dnses() {
     exit 1
   fi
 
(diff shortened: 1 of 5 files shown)


hack/eks/install.sh
```diff
@@ -77,7 +77,7 @@ capact::aws::install::capact() {
   shout "Deploying Capact..."
   capact install --environment eks \
     --version "${CAPACT_VERSION}" \
-    --capact-overrides "global.domainName=${CAPACT_DOMAIN_NAME}" \
+    --capact-overrides "global.domainName=${CAPACT_DOMAIN_NAME},gateway.ingress.annotations.class=capact" \
```
mszostok, line 80: I can also move it to CLI if needed.

pkosiec: We can keep it as it is 👍

pkosiec (approved):
👌

PR #568, opened 2021-11-24
Support GET requests and token auth in GitLab runner

pkosiec:
## Description

- Fix GitLab API Runner panic when provided body is empty
- Fix panic when created request is nil
- Support GET requests with query parameters
- Support personal token authentication in GitLab API Runner

## Testing

You can test it in isolation with the following input to the GitLab runner (see the [instruction](https://github.com/capactio/capact/blob/main/cmd/gitlab-api-runner/README.md):

```yaml
method: GET
path: "projects"
baseURL: "{baseURL}"
auth:
  token: "{token}"
body: null
queryParameters:
  simple: ["true"]
  search: ["{repo-path}"]
output:
  goTemplate: |
    id: {{ (index . 0).id }}
```

### DIFF
--- a/pkg/runner/gitlab-api-runner/runner.go
+++ b/pkg/runner/gitlab-api-runner/runner.go
@@ -31,12 +31,7 @@ func (r *RESTRunner) Do(ctx context.Context, in runner.StartInput) (*runner.Wait
 		opts = append(opts, gitlab.WithBaseURL(input.Args.BaseURL))
 	}
 
-	// TODO: add switch based on auth type
-	git, err := gitlab.NewBasicAuthClient(
-		input.Args.Auth.Basic.Username,
-		input.Args.Auth.Basic.Password,
-		opts...,
-	)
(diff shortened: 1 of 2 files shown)


pkg/runner/gitlab-api-runner/runner.go
```diff
@@ -104,7 +104,29 @@ func (r *RESTRunner) readInputData(in runner.StartInput) (Input, error) {
...
+	// basic credentials
+	if auth.Basic != nil {
+		return gitlab.NewBasicAuthClient(
+			auth.Basic.Username,
+			auth.Basic.Password,
+			opts...,
+		)
+	}
+
+	return nil, errors.New("no token or basic credentials provided")
```
mszostok, line 126: IMO it will be good to return an error when both basic and token are provided, this may reduce a lot of time spent on debugging in case of some misconfiguration :D

mszostok (approved):
LGTM 👍 just one comment, please address it before merging, thanks! 🙇

mszostok (approved):

PR #572, opened 2021-11-29
Enable CORS for Capact Gateway Ingress

pkosiec:
## Description

- Enable CORS for Capact Gateway Ingress

It enables Gateway GraphQL API to be consumed by UIs deployed anywhere by default only for local installation.

- [x] To discuss, which default values should we set: should we enable CORS by default and allow all origins? Or keep such overrides for local Capact for convenience if someone wants to develop UI for it?


mszostok (commented):
Personally, I would vote to allow all origins only in local development. It can be easily added to our CLI overrides: https://github.com/capactio/capact/blob/3ba7ad153e5afc9b99659b70f14df3536baabc52/internal/cli/capact/defaults.go#L58-L61

mszostok (approved):

PR #573, opened 2021-12-02
Remove unresolved fields of input and output TypeInstances from Engine APIs

pkosiec:
## Description

- Remove unresolved fields of input and output TypeInstances from Engine APIs:
     - name for output TypeInstance
     - typeRef for input TypeInstances

Unfortunately, it would be time-consuming to fill these fields properly as it would be a part of rendering process. As no one uses them, I think we can remove them safely.

## Testing

1. Run Capact from the PR (`make dev-cluster`)
2. Rebuild Capact CLI (`make build-tool-cli && mv ./bin/capact-darwin-amd64 /usr/local/bin/capact`)
2. Create Action with the [`typeinstance` greet example](https://github.com/capactio/hub-manifests/tree/main/examples/typeinstance)
2. Test Action query from GraphQL examples and observe updated API


pkg/engine/k8s/api/v1alpha1/action_types.go
```diff
@@ -320,6 +320,18 @@ func (r *RenderingStatus) SetInputParameters(params []byte) {
 	r.Input.SetParameters(params)
 }
 
+// SetInputTypInstances sets the Action input TypeInstances.
+func (r *RenderingStatus) SetInputTypInstances(typeInstances []InputTypeInstance) {
```
mszostok, line 324: It's not directly connected with this, but please remove `CURRENTLY NOT IMPLEMENTED.` from https://github.com/capactio/capact/blob/de21b30a416a4d63152865d8bb713da516de6542/pkg/engine/k8s/api/v1alpha1/action_types.go#L278. Thanks!

mszostok (approved):
lgtm 👍

PR #592, opened 2021-12-23
Ensure map initialized before accessing it, create secret only if needed

mszostok:
## Description

- Ensure that map is initialized before accessing it
- Create Secret by gql service only if needed
- Ensure that Secrets is owned by Action

### Testing

Check bug description https://github.com/capactio/capact/issues/593

## Related issue(s)

Fix https://github.com/capactio/capact/issues/593


pkosiec (approved):
LGTM

Issue #600, opened 2022-01-17
Release Capact 0.6.0


pkosiec:
## Description

 - [x] Create new GitHub release according to the https://capact.io/community/release doc.
     - Make sure we use just the images that are built on the main/release branches. Do not use `PR-XXX` tags
     - [x] Write release notes on top of the changelog (different section) and make sure they are approved by reviewers
     - Consider a super mini blog post about the release:
         - @pkosiec: Unfortunately, I ran out of time, so I wasn't able to write such blog post.
 - [x] Share the news on social media
 - [x] Take a final look at our Mattermost and RocketChat manifests and verify if they run properly on a fresh VM
         - Tested on https://github.com/capactio/capact/commit/ab6a8b3e86f9fa055feedc1f379c81fb92ef9cbe
 - [x] Nice to have: For dashboard set up automatic job to create branch + tag / manually if we don't have time
         -  capactio/dashboard#20
         -  capactio/capact#617

## Reason

We're about to finish our next milestone 🚀

pkosiec:
Here is a preview for new release description:

```
🎉 **Capact v0.6.0** is finally here! We try to keep up the pace to deliver brand-new features and other improvements to you. So, what's new, you ask?

### ✨ Key highlights

- Introducing Capact Dashboard - a brand-new management UI for Capact! With the initial version, you can explore Public Hub and manage your Actions. Read more [here](https://capact.io/docs/dashboard-ui/overview)!
- Our manifest generation tooling is out of alpha! Apart from supporting all kinds of OCF manifests, now you can use a very convenient, interactive command to generate multiple kind of manifests. To learn more, see the [`capact manifest generate`](https://capact.io/docs/cli/commands/capact_manifest_generate) docs.
- Populator for Capact Public Hub now supports multiple source repositories! No need to fork Capact Hub Manifests repository and add your own manifests on top on that. Check out the updated [Manage Public Hub manifests](https://capact.io/docs/example/public-hub-content) document.
- New manifests and runners: install Redis and MinIO using Capact! Also, the new [simple "Hello Capact" manifests](https://github.com/capactio/hub-manifests/tree/main/examples) make easy to learn what is Capact. Stay tuned for a blog post regarding this topic!
- Other bug fixes and improvements, such as:
    - enhanced Capact installation stability,
    - improved JSON schema for many manifests,
    - updated dependencies,
    - manifest validation enhancements,
    - and many, many more!
```

mszostok:
[...]

PR #617, opened 2022-01-26
Allow setting Dashboard image tag during Capact release

pkosiec:
## Description

- Allow setting Dashboard image tag during Capact release
- Refactor make-release.sh script to use `yq` (as previous approach with `sed` was error prone)

## Testing

Sample run: https://github.com/pkosiec/capact/actions/runs/1751316086

(I commented unnecessary steps: https://github.com/pkosiec/capact/commit/71cc8720651706b8e6bdd262d7ba2ab7d4313725)

Results:
- https://github.com/pkosiec/capact/commit/8471202432564c1488bf0cc43263d96514a7ddb1
- https://github.com/pkosiec/capact/commit/7b434e39dc357bf2371c78dc839554e1753d36c0

## Related issue(s)

https://github.com/capactio/capact/issues/600

### DIFF
--- a/.github/workflows/branch-build.yaml
+++ b/.github/workflows/branch-build.yaml
@@ -140,8 +140,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2 # To be able to compare with the previous commit and detect changed files
-      - name: Install yq
-        run: sudo snap install yq --channel=v4/stable
       - uses: google-github-actions/setup-gcloud@master
         with:
           service_account_key: ${{ secrets.CAPACT_GCS_CREDS }}
--- a/.github/workflows/branch-build.yaml
+++ b/.github/workflows/branch-build.yaml
@@ -140,6 +140,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2 # To be able to compare with the previous commit and detect changed files
+      - name: Install yq
+        run: sudo snap install yq --channel=v4/stable
       - uses: google-github-actions/setup-gcloud@master
         with:
           service_account_key: ${{ secrets.CAPACT_GCS_CREDS }}
--- a/.github/workflows/make-release.yaml
+++ b/.github/workflows/make-release.yaml
@@ -46,8 +46,6 @@ jobs:
           go-version: ${{env.GO_VERSION}}
       - name: Set up GoReleaser
         run: go install github.com/goreleaser/goreleaser@v1.1.0
-      - name: Install yq
-        run: sudo snap install yq --channel=v4/stable
       - name: Set up GCS
         uses: google-github-actions/setup-gcloud@master
         with:
--- a/.github/workflows/make-release.yaml
+++ b/.github/workflows/make-release.yaml
@@ -46,6 +46,8 @@ jobs:
           go-version: ${{env.GO_VERSION}}
       - name: Set up GoReleaser
         run: go install github.com/goreleaser/goreleaser@v1.1.0
+      - name: Install yq
+        run: sudo snap install yq --channel=v4/stable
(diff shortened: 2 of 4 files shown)


mszostok (approved):
I tested that and overall it works, but the only thing that I don't like about it is that it breaks our formatting: https://github.com/mikefarah/yq/issues/465 and produces less readable `vaules.yaml` files.

**Personally**, I'm not sure whether this change is really needed. The current solution worked IMO better. I know that it was error-prone, but this is not a user facing script. 

Maybe we should just have our own tool to replace proper things in our manifests.

pkosiec:
While I understand that `yq` reformats yamls removing e.g. blank lines, it still keep comments and other values, which doesn't seem to me as a huge downside. So I have to disagree with

> The current solution worked IMO better. I know that it was error-prone, but this is not a user facing script.

The current solution is very shortsighted. In case of Dashboard tag, it forces me to replace any `tag` property, which is a generic word. It's a matter of time when we have such duplicated property in the `values.yaml` file. Same with other properties, like `branch`, which could occur multiple times, as we have multiple sources support in Populator. Should we build more complex `sed` one-liners in future? Definitely not.

> Maybe we should just have our own tool to replace proper things in our manifests.

And this sounds like an overkill for me, but of course you can contribute with such solution 🙂  I backed off with the `yq` refactor (reverted modifications in two files) to end discussion, as I need the new functionality for upcoming release.

So, here are examples after changes:
- Failed run (image with `PR-` prefix`): https://github.com/pkosiec/capact/runs/4953623108?check_suite_focus=true
- Successful run: https://github.com/pkosiec/capact/runs/4953647640?check_suite_focus=true
- Results:
    - https://github.com/pkosiec/capact/commit/eb9de8f0efa3ef449b86856bb3821c66730c3557
    - https://github.com/pkosiec/capact/commit/738bbb4d4d55a03e5cda1400cfff7854d7ed4387 

(After that I fixed linter issues, but I also tested the script locally, it works the same).

mszostok (approved):
I totally get your concerns and I'm not against fixing that at all 👍 
But I think that it will be good to just log an issue for it instead of doing that here. We should aim for a solution which will solves the problem that you described but also preserve current good formatting approach 👍 

>And this sounds like an overkill for me,

I was referring to rewriting bash to Go which IMO is not so overkill at all as we are quite fluent in Go. Having:
```bash
# DASHBOARD_IMAGE_TAG - Dashboard image tag used for a given release
[[ ( -z "${DASHBOARD_IMAGE_TAG}" || "${DASHBOARD_IMAGE_TAG}" == PR-* ) ]] && echo "Need to set DASHBOARD_IMAGE_TAG that doesn't start with 'PR-' prefix" && exit 1;

SOURCE_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
RELEASE_VERSION_MAJOR_MINOR="$(echo "${RELEASE_VERSION}" | sed -E 's/([0-9]+\.[0-9])\.[0-9]/\1/g')"
RELEASE_BRANCH="release-${RELEASE_VERSION_MAJOR_MINOR}"
```
is definitely not our thing.

> And this sounds like an overkill for me, but of course you can contribute with such solution 🙂 

As mentioned, we can log an issue for that.

Regarding https://github.com/mikefarah/yq/issues/465, it was not working like that previously, so definitely there will be a solution for that.

TL;DR;
I'm totally not against improvements, and you spot a real problem but let's not do them in a rush where we solve one problem but introduce another one. If you can, please create a ticket for it, so we can come up with a solution and introduce it also in all repos at once. As we all known "consistent code is easier to maintain, and requires less cognitive overhead."

BTW previous solution was also approved, but as you revert it I also approve this one 👍

PR #619, opened 2022-01-27
Fix vulnerability detected by dependabot

mszostok:
## Description

- Fix vulnerability detected by dependabot

https://github.com/capactio/capact/security/dependabot/hub-js/package-lock.json/node-fetch/open

I bumped the express server to use the new node-fetch version. The `neo4j-graphql-js@2.19.4` is deprecated and we are not able to bump it higher, so it reports the problem that the version of `node-fetch` doesn't match but still works properly.


mkuziemko (approved):
:+1:

Issue #624, opened 2022-01-31
Update Policy API and PolicyEnforcedClient


mszostok:
## Description

Update Policy syntax and related logic according to the [Delegated Storage proposal](https://github.com/capactio/capact/blob/main/docs/proposal/20211207-delegated-storage.md).

## Syntax

- Breaking change:
     <table>
     <tr>
     <td> Before </td> <td> After </td>
     </tr>
     <tr>
     <td>

    ```yaml
    rules:
    - interface:
        path: cap.interface.database.postgresql.install
        revision: 0.1.0 # exact revision
      oneOf:
        - implementationConstraints:
            # (...)
    ```

     </td>
     <td>

    ```yaml
    interface:
      rules:  # rules for Interfaces, now nested under `interface`
      - interface:
          path: cap.interface.database.postgresql.install
          revision: 0.1.0 # exact revision
        oneOf:
          - implementationConstraints:
              # (...)
    ```

     </td>
     </tr>
     </table>

- Enhancements:
   - TypeInstances policy:
    ```yaml
    # ...
    
    typeInstance:
      rules:
        - typeRef:
            path: "cap.type.aws.auth.credentials"
            revision: "0.1.0" # optional
          backend:
            id: "00fd161c-01bd-47a6-9872-47490e11f996" # Vault backend storage
        - typeRef:
            path: "cap.type.aws.*" # for any Type reference starting with such prefix
          backend:
            id: "31bb8355-10d7-49ce-a739-4554d8a40b63" # AWS secrets manager
        - typeRef:
            path: "cap.*" # Any other Type reference
          backend:
            id: a36ed738-dfe7-45ec-acd1-8e44e8db893b
            description: "Default Capact PostgreSQL backend"
    ```
    if `capact-outputTypeInstances[*].backend` not specified, check default from Policy. 
    
    ```yaml
    # default - static
    capact-outputTypeInstances:
      - name: mattermost-config
        from: additional
        # no backend definition -> use default storage backend
    ```
    
    During render:
    ```yaml
    typeInstances:
    - alias: helm-release
      attributes: []
      createdBy: default/act-l49vh-30c7a078-6a77-475c-94dd-7466f56447ce
      typeRef:
        path: cap.type.helm.chart.release
        revision: 0.1.0
      value: null
      backend:
        id: 3ef2e4ac-9070-4093-a3ce-142139fd4a16 # helm-release backend - resolved UUID based on the injected TypeInstance
    ```

[...]

PR #625, opened 2022-01-31
Nest Interface Policy under the `interface` field

mszostok:
## Description

- Nest Interface Policy under the `interface` field
- Use our Go client to update Policy instead of doing that on raw ConfigMap. The config map update was implemented before we had the GraphQL cli: https://github.com/capactio/capact/blob/05383cd8fe0913242816316bab20bf21d445e43a/test/e2e/action_test.go#L266 Currently, it doesn't make sens to use it. It's better to test our client and do that via GraphQL to simulate a real user interaction.
  - For now I decided, to leave it as Global Policy, but IMO it will be good to change it to Action Policy, so the global cluster state won't be changed during executing e2e tests.

NOTE: The easiest way to review is to select **"Hide whitespaces"**.
Screen Shot 2022-01-31 at 14 34 05

## Related issue(s)

- https://github.com/capactio/capact/issues/624

### DIFF
--- a/deploy/kubernetes/charts/capact/charts/engine/values.yaml
+++ b/deploy/kubernetes/charts/capact/charts/engine/values.yaml
@@ -60,67 +60,57 @@ affinity: {}
 
 # order from highest priority to the lowest
 policyOrder: "ACTION,GLOBAL,WORKFLOW"
-globalPolicyRules:
+globalPolicy:
 # Insert Interface paths with Implementations. For example:
-#
-#  - interface:
-#      # Rules for exact path and revision (looked up in first place)
-#      path: "cap.interface.database.postgresql.install"
-#      revision: "0.1.0"
-#    oneOf:
-#      - implementationConstraints:
-#          requires:
-#            - path: "cap.type.gcp.auth.service-account"
-#              # any revision
-#          attributes:
-#            - path: "cap.attribute.cloud.provider.gcp"
-#              # any revision
-#        inject:
-#         typeInstances:
-#           - id: "{uuid}"
-#             typeRef:
-#               path: "cap.type.gcp.auth.service-account"
-#               revision: "0.1.0"
-#      - implementationConstraints:
-#          attributes:
-#           - path: cap.attribute.cloud.provider.aws
-#             # any revision
-#      - implementationConstraints:
-#          path: cap.implementation.bitnami.postgresql.install
-#  - interface:
-#      # Rules for exact path in any revision (looked up in second place)
-#      path: "cap.interface.database.postgresql.install"
-#    oneOf:
-#      - implementationConstraints:
-#          requires:
-#            - path: "cap.type.gcp.auth.service-account"
-#              # any revision
-#          attributes:
-#            - path: "cap.attribute.cloud.provider.gcp"
-#              # any revision
+#  interface:
+#    rules:
+#      - interface:
+#          # Rules for exact path and revision (looked up in first place)
+#          path: "cap.interface.database.postgresql.install"
+#          revision: "0.1.0"
+#        oneOf:
+#          - implementationConstraints:
+#              requires:
+#                - path: "cap.type.gcp.auth.service-account"
+#                  # any revision
+#              attributes:
+#                - path: "cap.attribute.cloud.provider.gcp"
+#                  # any revision
+#            inject:
+#             typeInstances:
+#               - id: "{uuid}"
+#                 typeRef:
+#                   path: "cap.type.gcp.auth.service-account"
+#                   revision: "0.1.0"
+#          - implementationConstraints:
+#              attributes:
+#               - path: cap.attribute.cloud.provider.aws
+#                 # any revision
+#          - implementationConstraints:
+#              path: cap.implementation.bitnami.postgresql.install
+#      - interface:
+#          # Rules for exact path in any revision (looked up in second place)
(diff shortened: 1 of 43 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -26,30 +29,44 @@ type GatewayConfig struct {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	input := enginegraphql.PolicyInput{}
+	err = yaml.Unmarshal(rawPolicy, &input)
+	Expect(err).ToNot(HaveOccurred())
+	originalGlobalPolicy = input
+
+	//waitTillServiceEndpointsAreReady()
+	//waitTillDataIsPopulated()
+})
```
mszostok, line 64: will be reverted before merge (after review)

internal/k8s-engine/graphql/domain/policy/converter.go
```diff
@@ -17,12 +17,26 @@ func NewConverter() *Converter {
...
+	if in == nil {
+		return policy.InterfacePolicy{}, nil
+	}
 	var rules policy.RulesList
 
 	for _, gqlRule := range in.Rules {
 		policyRules, err := c.policyRulesFromGraphQLInput(gqlRule.OneOf)
 		if err != nil {
-			return policy.Policy{}, errors.Wrap(err, "while getting Policy rules")
+			return policy.InterfacePolicy{}, err
```
mkuziemko, line 39: we should continue to use `errors.Wrap` in this file

deploy/kubernetes/charts/capact/charts/engine/values.yaml
```diff
@@ -60,67 +60,57 @@ affinity: {}
...
+          # any other Interface (looked up in third place, if there is no rule for `path` and `revision` or `path`)
+          path: "cap.*"
+        oneOf:
+          - implementationConstraints: # prefer Implementation for Kubernetes
+              requires:
+                - path: "cap.core.type.platform.kubernetes"
+                  # any revision
+          - implementationConstraints: {} # fallback to any Implementation
+
+## Currently, not used but setting this filed is supported by all scripts and CLI.
```
mkuziemko, line 114: If we comment out this value from the Chart, would it work:
```
// from hack/lib
CAPACT_OVERRIDES+=",engine.testSetup.enabled=${USE_TEST_SETUP}"
```
? 
I didn't check it but I am wondering if it is ok to set up a value that does not exist in the Helm chart.

pkg/engine/k8s/policy/metadata/metadata.go
```diff
@@ -47,7 +47,9 @@ func (m TypeInstanceMetadata) String(withKind bool) string {
 // TypeInstanceIDsWithUnresolvedMetadataForPolicy filters TypeInstances that have unresolved metadata.
 func TypeInstanceIDsWithUnresolvedMetadataForPolicy(in policy.Policy) []TypeInstanceMetadata {
 	var tis []TypeInstanceMetadata
-	for _, rule := range in.Rules {
+
+	// Interface
```
mkuziemko, line 51: Not needed

test/e2e/action_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
mkuziemko, line 1: for curiosity, why do we need this line, and what it does do? Is it something different than `+build integration`

mkuziemko (commented):
Good work :+1: Some suggestions/questions added.

deploy/kubernetes/charts/capact/charts/engine/values.yaml
mszostok: in `values.yaml` file you store only default that you want to have. Still others may specify it and chart template may use it.
Screen Shot 2022-02-04 at 09 17 44
 
 
 As we don't have https://helm.sh/docs/topics/charts/#schema-files it won't report any error.

pkg/engine/k8s/policy/metadata/metadata.go
mszostok: I added that, to show that only one part is currently processed. In next PR I will add support for TypeInstances.

Screen Shot 2022-02-04 at 09 21 34


I added that as we use  “grouping” comments in other func in this file: https://github.com/capactio/capact/blob/49defde437eb06f6ff249aa055646f52c52836f7/pkg/engine/k8s/policy/metadata/metadata.go#L67-L92

test/e2e/action_test.go
mszostok: Because in 1.17 they changed the syntax for build tags. The `go fmt` adds both of them to gives you an option to easily migrate your code.

I recommend you to take a look at:
- https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md
- or recording https://www.youtube.com/watch?v=AgR_mdC4Rs4
in a free time, to learn why they changed it.

internal/k8s-engine/graphql/domain/policy/converter.go
mszostok: you're right, fixed 👍

test/e2e/action_test.go
mkuziemko: Thanks, I will check :+1:

mkuziemko (approved):
LGTM :+1:

Issue #656, opened 2022-03-09
Add Storage TypeInstance related validation


mszostok:
## Description

We need to ensure that all operation related to Storage TypeInstance are valid. 

## AC

- validate if backend accepts `value` field
- validate if backend accepts `context` field
  - if accepts, validate data against `backend.contextSchema`
- validate that Storage TypeInstances follow the convention of having `url`, `acceptValue`, and `contextSchema` fields. As context is optional, the `contextSchema` field is nullable.

out of scope:
- validate TypeInstance's `value` field. This is not directly related to TypeInstance Storage and will be handled by https://github.com/capactio/capact/issues/633

## Related issues

See epic #604 for reason and use cases.

PR #659, opened 2022-03-11
Add Storage TypeInstance related validation

mszostok:
## Description

- validate if backend accepts `value` field
- validate if backend accepts `context` field
  - if accepts, validate data against `backend.contextSchema`
- validate that Storage TypeInstances follow the convention of having `url`, `acceptValue`, and `contextSchema` fields. As context is optional, the `contextSchema` field is nullable.

It was done directly in each related method, I tried to create a dedicated middleware (draft code: https://github.com/capactio/capact/commit/1273fc332f2c08f5b739374a31456b5421e3f9ef) but it didn't work well:
- problems with decoding - the request is already encoded
- problems with strong typing
- more complicated setup than the inlined version

## Testing

The test that I wrote, proves that it works e2e (Go client -> GraphQL -> Local Hub -> External Storage).

1. Run neo4j:
    ```bash
    docker run -d \
      -p 7687:7687 -p 7474:7474 \
      -e "NEO4J_AUTH=neo4j/okon" \
      -e "NEO4JLABS_PLUGINS=[\"apoc\"]" \
      --name hub-neo4j-instance \
      ghcr.io/capactio/neo4j:4.2.13-apoc
    ```

2. Run Local Hub:
    ```bash
    cd hub-js; APP_NEO4J_ENDPOINT=bolt://localhost:7687 APP_NEO4J_PASSWORD=okon APP_HUB_MODE=local npm run dev; cd ..
    ```

3. Run test:
    ```bash
    RUN_VALIDATION=true go test ./pkg/hub/client/local/ -v -count 1
    ```

## Questions

1. Currently, there is an early return in case of error, so if both `value` and `context` are incorrect only information about `value` is returned:
    ```bash
    while executing mutation to create TypeInstance: All attempts fail:
    #1: graphql: failed to create TypeInstance: failed to create the TypeInstances: 2 error occurred:
      * Error: External backend "MOCKED_ID": input value not allowed
    ```
    Is it ok?  if not, we will need to support throwing multierr to print aggregated data in higher layer.
    
2. As you may notice in test the errors are duplicated:
    
    ```bash
    while executing mutation to create TypeInstance: All attempts fail:
    #1: graphql: failed to create TypeInstance: failed to create the TypeInstances: 2 error occurred:
      * Error: External backend "MOCKED_ID": input value not allowed
      * Error: rollback externally stored values: External backend "MOCKED_ID": input value not allowed`),
    ```
    but they are related to different things. The first one is about creating an obj in delegated storage and the second one is about doing a rollback as the error was raised. Currently, the logic is fairly simple, if there was an error in transaction we ensure that the data was removed from external storage, and it doesn't need to be always the same error, so IMO it is ok to return both of them.
## Related issue(s)

Fix https://github.com/capactio/capact/issues/656


mszostok:
e2e failed but it will be fixed by: https://github.com/capactio/capact/pull/660

pkg/hub/client/local/client_test.go
```diff
@@ -0,0 +1,274 @@
...
+				URL:         ptr.String("http://localhost:5000/fake"),
+				AcceptValue: ptr.Bool(false),
+			},
+			value: map[string]interface{}{
+				"name": "Luke Skywalker",
+			},
+			context: map[string]interface{}{
+				"name": "Luke Skywalker",
+			},
+			// TODO(review): currently it's a an early return, is it sufficient?
```
mkuziemko, line 144: In my opinion, yes :+1:

pkg/hub/client/local/client_test.go
```diff
@@ -0,0 +1,274 @@
...
+				AcceptValue: ptr.Bool(false),
+			},
+			value: map[string]interface{}{
+				"name": "Luke Skywalker",
+			},
+			expErrMsg: heredoc.Doc(`
+              while executing mutation to create TypeInstance: All attempts fail:
+              #1: graphql: failed to create TypeInstance: failed to create the TypeInstances: 2 error occurred:
+              	* Error: External backend "MOCKED_ID": input value not allowed
+              	* Error: rollback externally stored values: External backend "MOCKED_ID": input value not allowed`),
```
mkuziemko, line 67: As I understand, one of the errors is from `Store` and the second is from `rollbackExternalStoreAction`.  Is it an option to not perform a rollback in case of a validation error?

pkg/hub/client/local/client_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+	"capact.io/capact/internal/cli/heredoc"
+	"capact.io/capact/internal/ptr"
+	gqllocalapi "capact.io/capact/pkg/hub/api/graphql/local"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO(review): THIS FILE WILL BE REMOVED BEFORE MERGING. IT WAS ADDED ONLY FOR DEMO/PR TESTING PURPOSES.
```
mkuziemko, line 18: I know that is only for testing purposes of the PR and we don't want to keep it in the repository but on the other hand, seems valid and covers a lot of cases. It could be handy when someone would like to change something in the validation error and want to make sure that regression is fine.

mkuziemko (commented):
Works well. Small questions/comments added.

pkg/hub/client/local/client_test.go
mszostok: Fixed 👍

pkg/hub/client/local/client_test.go
mszostok: I tried to but because we can create in a single mutation more than one TypeInstance it complicates the solution, so I decide to give up.

mkuziemko (approved):
Nice work! :rocket:

PR #662, opened 2022-03-14
Fix resolving backend.context in cypher

mszostok:
## Description

- Fix resolving `backend.context` in cypher. Currently, the context was always taken from the user input even if the `backend` was not sent. This against our contract, in case of empty backend we should use the old `backend.context`

Update scenario:
- if `backend` is not sent, use old settings
- if `backend` was sent, use new `backend.context` value (if it's `null` then persist `null` which means, zero `backend.context` property)

## Testing

1. Run neo4j:
    ```bash
    docker run -d \
      -p 7687:7687 -p 7474:7474 \
      -e "NEO4J_AUTH=neo4j/okon" \
      -e "NEO4JLABS_PLUGINS=[\"apoc\"]" \
      --name hub-neo4j-instance \
      ghcr.io/capactio/neo4j:4.2.13-apoc
    ```
2. Run storage server:
    ```bash
    APP_LOGGER_LEVEL="debug" APP_LOGGER_DEV_MODE=true APP_SUPPORTED_PROVIDERS="dotenv" go run ./cmd/secret-storage-backend/main.go
    ```
3. Run Local Hub:
    ```bash
    cd hub-js; APP_NEO4J_ENDPOINT=bolt://localhost:7687 APP_NEO4J_PASSWORD=okon APP_HUB_MODE=local npm run dev; cd ..
    ```

5. Run test:
    ```bash
    GRPC_SECRET_STORAGE_BACKEND_ADDR="0.0.0.0:50051" go test ./pkg/hub/client/local/ -v -count 1
    ```

You can also manually test that using GraphQL playground.


hub-js/graphql/local/schema.graphql
```diff
@@ -503,7 +503,9 @@ type Mutation {
           RETURN specBackend
         ',
         '
-          CREATE (specBackend: TypeInstanceResourceVersionSpecBackend {context: apoc.convert.toJson(item.typeInstance.backend.context)})
+          MATCH (latestRevision)-[:SPECIFIED_BY]->(latestSpec: TypeInstanceResourceVersionSpec)
```
pkosiec, line 506: While it works for:

✅  Case 1: Set context to null
```graphql
mutation UpdateTypeInstances($typeInstanceId: ID!) {
  updateTypeInstances(
    in: [
      {
        ownerID: "namespace/name"
        id: $typeInstanceId
        typeInstance: {
          attributes: [{ path: "cap.attribute.sample", revision: "0.1.0" }]
          value: { name: "Luke Skywalker3" }
          backend: {
            context: null 
          }
        }
      }
    ]
  ) {
    id
    
  }
}
```
and

✅  Case 2: Do not provide backend - reuse old context
```graphql
mutation UpdateTypeInstances($typeInstanceId: ID!) {
  updateTypeInstances(
    in: [
      {
        ownerID: "namespace/name"
        id: $typeInstanceId
        typeInstance: {
          attributes: [{ path: "cap.attribute.sample", revision: "0.1.0" }]
          value: { name: "Luke Skywalker3" }
         # no backend at all - reuse old context
        }
      }
    ]
  ) {
    id
    
  }
}

```

I think it should also work for:

❌  Case 3: Provide backend object, but without specifying `context` field:

```graphql
mutation UpdateTypeInstances($typeInstanceId: ID!) {
  updateTypeInstances(
    in: [
      {
        ownerID: "namespace/name"
        id: $typeInstanceId
        typeInstance: {
          attributes: [{ path: "cap.attribute.sample", revision: "0.1.0" }]
          value: { name: "Luke Skywalker3" }
          backend: { 
            # context: {
            #  provider: "foo"
            # }
          }
        }
      }
    ]
  ) {
    id
    
  }
}
```

But it doesn't - it makes the context `null`. It would make sense, cause there might be more fields inside `backend` object, and if I don't want to change `context`, I just don't provide it.

In JS it is easy to check whether a given property is `null` or `undefined` (not provided by user). Not sure if we can achieve this in Cypher though 🤔  If that's not easy, let's skip it 🙂

pkosiec (approved):
LGTM with one improvement suggestion. If that's not something easy to do, let's skip it and merge this PR 👍

hub-js/graphql/local/schema.graphql
mszostok: Yep, there is a test that showcase this scenario, unfortunately AFAIK it's not possible with Cypher (https://neo4j.com/developer/kb/understanding-non-existent-properties-and-null-values/). Probably this will need to be done on the JS side but we use also `cypherMutation` which generates this query and due to that fact it's not easy. 

As we don't have any other field under `backend` I will go with what you proposed - skip it.
We can revisit that when a new property will be added,  and figure out how to handle that nicely. But IMO it won't be done in the near future 😄

PR #669, opened 2022-03-16
Bootstrap Helm storage backend Go app

pkosiec:
## Description

- Bootstrap Helm storage backend Go app for both Release and Templating gRPC servers

**Additional:**
- Add missing Go comments for macOS-only source code (to avoid lint errors on macOS)
- Fix duplicated logger name on healthz endpoints
- Make new Hadolint not failing on syntax comment in Dockerfile

## Testing

No testing required, as the services are not yet implemented. But you can run them anyway if you want - just follow the README.md inside `cmd/helm-storage-backend/` directory 🙂 

You can also run some gRPC calls from Insomnia, but unfortunately the response won't be that useful - that's why I added logging to the sample `GetValue` method.
Screenshot 2022-03-16 at 15 56 58

## Related issue(s)

#670

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# hadolint ignore=DL3061
 # syntax = docker/dockerfile:1-experimental
 
 FROM golang:1.16.3-alpine as builder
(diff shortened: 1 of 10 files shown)


mszostok (approved):
LGTM 🚀

Issue #670, opened 2022-03-16
Setup Helm storage code and manifests


pkosiec:
## Description

The most important thing is that the point no 1 needs to be done ASAP to unblock other subtasks.

- [x] Bootstrap Go code for the helm storage backends
    - Single entrypoint (directory in `cmd`)
    - Based on flags, a different storage backend gRPC server is run
    - Both servers (release + template) implement the same Storage Backend ProtoBuf interface (use already generated Go code)
    - Modify pipelines to build a single Docker image
- [ ] Add dedicated Helm chart and Hub manifests to install Helm storage backends
    - [x] Single Helm chart, one deployment, two containers and two separate services
    - [x] Update pipelines for releasing Helm charts
    - The implementation produces two TypeInstances for two storage backends (helm release backend + helm template backend) - define them
    - Write instruction on PR how to install such storage backend. It will be later reused in documentation
    - Base on secret storage backend, as this is pretty similar case (https://github.com/capactio/capact/issues/647)

## Related issues

See epic #650

pkosiec:
PRs:
- https://github.com/capactio/capact/pull/669
- https://github.com/capactio/capact/pull/675
- https://github.com/capactio/hub-manifests/pull/63

PR #675, opened 2022-03-16
Add Helm chart for Helm backend storage

pkosiec:
## Description

- Add Helm chart for Helm backend storage

## Testing

```bash
# Install chart (ofc you can use different name of the helm release)
helm install capact ./deploy/kubernetes/charts/helm-storage-backend --set=global.containerRegistry.path="ghcr.io/capactio/pr" --set=global.containerRegistry.overrideTag="PR-669" --wait

# Get logs
k logs capact-helm-storage-backend-{random} release -f
k logs capact-helm-storage-backend-{random} template -f

# Port forward
kubectl port-forward svc/capact-helm-storage-backend-release 3000:50051
# in second terminal:
kubectl port-forward svc/capact-helm-storage-backend-template 3001:50052
```

Now, you can do some gRPC calls via Insomnia to `localhost:3000` and `localhost:3001` and observe logs to see that proper backends are called.

Screenshot 2022-03-16 at 16 28 42

## Related issue(s)

#670

### DIFF
--- a/deploy/kubernetes/charts/helm-storage-backend/Chart.yaml
+++ b/deploy/kubernetes/charts/helm-storage-backend/Chart.yaml
@@ -1,0 +1,15 @@
+apiVersion: v2
+name: helm-storage-backend
+description: A Helm chart for Helm Storage Backend for Capact Local Hub
+
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.6.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 0.0.1
--- a/deploy/kubernetes/charts/helm-storage-backend/templates/serviceaccount.yaml
+++ b/deploy/kubernetes/charts/helm-storage-backend/templates/serviceaccount.yaml
@@ -1,12 +1,6 @@
-{{- if .Values.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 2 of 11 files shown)


deploy/kubernetes/charts/helm-storage-backend/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+name: helm-storage-backend
+description: A Helm chart for Helm Storage Backend for Capact Local Hub
+
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.6.0
```
mszostok, line 10: **nit:** in other places we have `0.1.0` as this won't be updated on the `main` branch. IMO for consistency it will be better to change it to `0.1.0`

deploy/kubernetes/charts/helm-storage-backend/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
```
mszostok, line 1: As those are Helm backends we will interact with K8s, so it will be also good to bootstrap cluster roles for them as they need to work across all Namespaces.

AFAIK we will need only:
- `get` and `list` on `configmaps`
- `get` and `list` on `secrets`

In case sth else will be needed, we can easily add it later if you will bootstrap them.

mszostok (approved):
LGTM 👍 but please add missing cluster role before merging this PR

deploy/kubernetes/charts/helm-storage-backend/Chart.yaml
pkosiec: This is not true: this version is updated while creating a new release. Please take a look at other charts inside `deploy/kubernetes/charts` dir.

However, this comment is true for every subchart (e.g. the ones under `capact` Helm chart).

PR #676, opened 2022-03-17
Add "safe-to-evict" autoscaler annotations for components with `emptyDir` mounted

pkosiec:
## Description

- Add "safe-to-evict" autoscaler annotations for components with `emptyDir` mounted

Prometheus, Alertmanager and Neo4j have `emptyDir` mounted which prevent them to scale down clusters that use Cluster Autoscaler, e.g. our long-running one on GCP:

Screenshot 2022-03-17 at 10 48 28

Since GKE 1.22 we wouldn't need such annotations on the pods:
https://cloud.google.com/kubernetes-engine/docs/release-notes#October_27_2021

## Related links
- https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node
- https://github.com/kubernetes/autoscaler/issues/966


mszostok (approved):
tested with `helm template`. LGTM 👍

Issue #671, opened 2022-03-16
Implement Helm release handler for Helm storage backend


pkosiec:
## Description

Implement Helm release storage backend:

- Extend Helm backend under `/cmd/helm-storage-backend`. Implement [gRPC server](https://github.com/capactio/capact/tree/main/pkg/hub/api/grpc/storage_backend.proto#L1). Types are already generated.
    - Support all possible releases backends (ConfigMap/Secret/etc). Maybe we will need to extend the `backend.context` to allow specifying it.
    - Support only `GetValue` method, the rest just validate data (e.g. onDelete checks whether a given Helm release exist).
        - To consider what exactly we want to validate - Should we really return error if a given Helm release doesn't exist in `OnDelete` method? It could be deleted externally. Or perhaps we just need to validate that in `onCreate`, `onUpdate`, and locking/unlocking? 🤔
- [Helm backend Type manifests is already defined](https://github.com/capactio/hub-manifests/blob/main/manifests/type/helm/storage.yaml). Ensure that it is up-to-date at the end of the task.
- Keep the same shape of data: https://github.com/capactio/hub-manifests/blob/main/manifests/type/helm/chart/release.yaml

## Related issues

See epic #650

PR #677, opened 2022-03-17
Implement Helm release storage backend

mszostok:
## Description

- Implement Helm release storage backend

Related PR: https://github.com/capactio/hub-manifests/pull/62. In backend, we have an assumption that the data is validated by Local Hub, so we receive a valid context with all required fields.

## Testing

**Setup**
1. Create k8s cluster. For example, run:
    ```bash
    k3d cluster create test
    ```
2. Install PostgreSQL Helm chart:
    ```bash
    RUNNER_CONTEXT_PATH=cmd/helm-runner/example-input/context.yaml \
     RUNNER_ARGS_PATH=cmd/helm-runner/example-input/install-args.yaml \
     RUNNER_LOGGER_DEV_MODE=true \
     RUNNER_COMMAND="install" \
     go run cmd/helm-runner/main.go
    ```
**Testing**
1. Run server:
    ```bash
    APP_LOGGER_DEV_MODE=true APP_MODE="release" go run ./cmd/helm-storage-backend/main.go
    ```
4. Run showcase test:
    ```bash
    GRPC_SECRET_STORAGE_BACKEND_ADDR=":50051" go test ./internal/helm-storage-backend/... -run "^TestShowcase$" -v -count 1
    ```

There is also unit test coverage. 
## Related issue(s)

Fix: https://github.com/capactio/capact/issues/671

### DIFF
--- a/internal/helm-storage-backend/common.go
+++ b/internal/helm-storage-backend/common.go
@@ -1,3 +1,0 @@
-package helmstoragebackend
-
-const defaultHelmDriver = "secrets"
--- a/internal/helm-storage-backend/common.go
+++ b/internal/helm-storage-backend/common.go
@@ -1,0 +1,3 @@
+package helmstoragebackend
+
+const defaultHelmDriver = "secrets"
--- a/internal/helm-storage-backend/release.go
+++ b/internal/helm-storage-backend/release.go
@@ -22,6 +22,7 @@ var _ pb.StorageBackendServer = &ReleaseHandler{}
 const latestRevisionIndicator = 0
 
 type (
+	// ReleaseDetails holds Helm release details.
 	ReleaseDetails struct {
 		// Name specifies installed Helm release name.
 		Name string `json:"name"`
@@ -30,6 +31,8 @@ type (
 		// Chart holds Helm Chart details.
 		Chart ChartDetails `json:"chart"`
 	}
+
+	// ChartDetails holds Helm chart details.
 	ChartDetails struct {
 		// Name specifies Helm Chart name.
 		Name string `json:"name"`
@@ -39,6 +42,7 @@ type (
 		Repo string `json:"repo"`
 	}
 
+	// ReleaseContext holds context used by Helm release storage backend.
 	ReleaseContext struct {
 		// Name specifies Helm release name for a given request.
 		Name string `json:"name"`
@@ -69,14 +73,17 @@ func NewReleaseHandler(log *zap.Logger, helmCfgFlags *genericclioptions.ConfigFl
 	}, nil
 }
 
+// OnCreate checks whether a given Helm release is accessible this storage backend.
 func (h *ReleaseHandler) OnCreate(_ context.Context, req *pb.OnCreateRequest) (*pb.OnCreateResponse, error) {
-	return &pb.OnCreateResponse{}, h.checkIfHelmReleaseExist(req.TypeInstanceId, req.Context)
+	if err := h.checkIfHelmReleaseExist(req.TypeInstanceId, req.Context); err != nil {
+		return nil, err
+	}
+
+	return &pb.OnCreateResponse{}, nil
 }
 
 // GetValue returns a value for a given TypeInstance.
 func (h *ReleaseHandler) GetValue(_ context.Context, req *pb.GetValueRequest) (*pb.GetValueResponse, error) {
-	h.log.Info("Getting value")
-
 	releaseContext, err := h.getReleaseContext(req.Context)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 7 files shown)


internal/helm-storage-backend/release.go
```diff
@@ -2,32 +2,206 @@ package helmstoragebackend
...
+	}
+
+	_, err = helmGet.Run(releaseContext.Name)
+	switch {
+	case err == nil:
+	case errors.Is(err, driver.ErrReleaseNotFound):
+		return status.Error(codes.NotFound, fmt.Sprintf("Helm release '%s/%s' for TypeInstance '%s' was not found", releaseContext.Namespace, releaseContext.Name, ti))
+	default:
+		return h.internalError(errors.Wrap(err, "while checking if Helm release exists"))
+	}
```
pkosiec, line 191: This is almost identical to the snippet from `GetValue`:

```go
	release, err := helmGet.Run(releaseContext.Name)
	switch {
	case err == nil:
	case errors.Is(err, driver.ErrReleaseNotFound):
		return nil, status.Error(codes.NotFound, fmt.Sprintf("Helm release '%s/%s' for TypeInstance '%s' was not found", releaseContext.Namespace, releaseContext.Name, req.TypeInstanceId))
	default:
		return nil, h.internalError(errors.Wrap(err, "while getting Helm release"))
	}
```

Consider having it defined only once, as it would be easier to maintain.

internal/helm-storage-backend/release_test.go
```diff
@@ -0,0 +1,497 @@
...
+			expectedDriver:      "configmaps",
+		},
+		{
+			name:                 "should ignore resourceVersion and return the latest release",
+			givenTypeInstanceID:  ptr.String("123"),
+			givenResourceVersion: 42, // should be ignored
+			expectedDriver:       "secrets",
+		},
+		{
+			name:           "should return the latest release even if TypeInstance's id and revision are not specified",
```
pkosiec, line 55: I'm not really sure if that's a useful test case - consider removing it. Our contract says that there should be TypeInstance ID, so I don't think we should even support this case. However, I'm not saying we should add an additional check for non-empty TypeInstance ID either, but I would simply remove this test case 😄

internal/helm-storage-backend/release_test.go
```diff
@@ -0,0 +1,497 @@
...
+	// globally given
+	const (
+		releaseName      = "test-release"
+		releaseNamespace = "test-namespace"
+	)
+	tests := []struct {
+		name string
+
+		request       *pb.OnCreateRequest
+		internalError error
```
pkosiec, line 235: This is not used anywhere in any of the test cases

internal/helm-storage-backend/release_test.go
```diff
@@ -0,0 +1,497 @@
...
+	// globally given
+	const (
+		releaseName      = "test-release"
+		releaseNamespace = "test-namespace"
+	)
+	tests := []struct {
+		name string
+
+		request       *pb.OnUpdateRequest
+		internalError error
```
pkosiec, line 338: Same as before - not used in test cases

pkosiec (approved):
Works well 👌  A few minor comments and this PR is ready to be merged 🙂

internal/helm-storage-backend/common.go
```diff
@@ -0,0 +1,3 @@
+package helmstoragebackend
+
+const defaultHelmDriver = "secrets"
```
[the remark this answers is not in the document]
mszostok: totally agree 👍

pkg/runner/cloudsql/create.go
```diff
@@ -174,7 +175,7 @@ func (a *createAction) createAdditionalOutputFile(path string, args *OutputArgs,
 		return errors.Wrap(err, "failed to load template")
 	}
 
-	fd, err := os.Create(path)
+	fd, err := os.Create(filepath.Clean(path))
```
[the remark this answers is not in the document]
mszostok: yep, for now I got an error on linter (locally I use a new version), so I decided to just fix it for now.

internal/helm-storage-backend/release_test.go
```diff
@@ -0,0 +1,497 @@
...
+		{
+			name:                "should use configmap driver and return the latest release",
+			givenTypeInstanceID: ptr.String("123"),
+			givenDriver:         ptr.String("configmaps"),
+			expectedDriver:      "configmaps",
+		},
+		{
+			name:                 "should ignore resourceVersion and return the latest release",
+			givenTypeInstanceID:  ptr.String("123"),
+			givenResourceVersion: 42, // should be ignored
```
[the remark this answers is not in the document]
mszostok: fixed

internal/helm-storage-backend/release_test.go
mszostok: fixed

internal/helm-storage-backend/release.go
mszostok: I thought about it, but it also complicates this function. However, I "merged" it as suggested.

PR #682, opened 2022-03-22
Update gcloud usage

mszostok:
## Description

- Fix problem with outdated gcloud usage, see https://github.com/capactio/capact/actions/runs/2019435422

## Note
As we don't push anything  on other pipeline steps, I also didn't use the logic for committing a `.gitignore` file.


pkosiec (approved):
LGTM - as discussed you can also prevent pushing to the origin, e.g. https://github.com/capactio/capact/blob/main/.github/workflows/make-release.yaml#L69-L82

Issue #665, opened 2022-03-15
Implement Go Template storage backend service


mszostok:
## Description

In Capact manifests, Content Developer is able to render `config` based on other TypeInstances data, see: 
https://github.com/capactio/hub-manifests/blob/9987550e2341cc761c8613cc4e6bb716d5a328f1/manifests/implementation/gcp/cloudsql/postgresql/install-0.2.0.yaml#L102-L108

This is not supported directly in new external backends. We want to have a dedicated backend which will do this projection.

Read more on [Delegated TypeInstance projection](https://github.com/capactio/capact/blob/main/docs/proposal/20211207-delegated-storage.md#dynamic-typeinstance-projections) in delegated proposal.

## AC
- Dedicated backend storage for TypeInstance projection created under `cmd/` in Capact repo.
- Support Go templates
- Backend dedicated documentation under https://capact.io/docs/next/feature/storage-backends/introduction

## Related issues

See epic #604 for reason and use cases.

PR #685, opened 2022-03-22
Add Helm template storage backend logic

mszostok:
## Description

- Add Helm template storage backend logic

## Testing

1. Create cluster:
    ```
    export DISABLE_MONITORING_INSTALLATION=true
    export ENABLE_POPULATOR=true
    export CLUSTER_TYPE=k3d
    export DISABLE_K3D_REGISTRY=false
    make dev-cluster
    ```
2. Install Helm backend via UI
3. Upgrade Helm backend release (new chart is not yet released):
    ```bash
    helm upgrade helm-storage-backend-1648070484 -n default ./deploy/kubernetes/charts/helm-storage-backend --set=global.containerRegistry.path="ghcr.io/capactio/pr" --set=global.containerRegistry.overrideTag="PR-685" --wait --reuse-values
    ```
     _Update helm release name._
4. Until https://github.com/capactio/capact/pull/657 is not merge, run:
    ```
    kubectl set image deploy/capact-hub-local hub-local=ghcr.io/capactio/pr/hub-js:PR-657
    ```
5. Install sample Helm release via Runner:
    ```bash
    RUNNER_CONTEXT_PATH=cmd/helm-runner/example-input/context.yaml \
     RUNNER_ARGS_PATH=cmd/helm-runner/example-input/install-args.yaml \
     RUNNER_LOGGER_DEV_MODE=true \
     RUNNER_COMMAND="install" \
     go run cmd/helm-runner/main.go
    ```
6. Get ID of  `cap.type.helm.template.storage:0.1.0`
7. Run test:
    ```bash
    BACKEND_ID="ecdfece6-da75-483f-a3ca-24246ed3308d" go test ./pkg/hub/client/ -run TestThatShowcaseExternalStorage -v -count 1
    ```

Unfortunately, in my case, it's quite slow. Probably we will need to debug it further. Currently, for my setup to work properly, it requires 1CPU.

The general problem is that the PostgreSQL has as a dependency “[common](https://github.com/bitnami/charts/blob/bda3812d402be83f4c68213961ce69ed22e26156/bitnami/postgresql/Chart.lock#L1-L4)” chart (small size). We need to download it (in general deps charts) as it won't be available by default. The [LocateChart method](https://github.com/helm/helm/blob/1ec0aacb8865d5b1f7ef1cb884bbf9b12579ecef/pkg/action/install.go#L675) from Helm project, that is able to download this chart, starts with [downloading an index file](https://github.com/helm/helm/blob/1ec0aacb8865d5b1f7ef1cb884bbf9b12579ecef/pkg/repo/chartrepo.go#L252) which in case of Bitnami has >3MB. This needs to be [decoded](https://github.com/helm/helm/blob/1ec0aacb8865d5b1f7ef1cb884bbf9b12579ecef/pkg/repo/index.go#L325-L338) and as a result it requires a much higher CPU limit.

You could say, that this is a special case, but unfortunately the PostgreSQL is quite often used in our examples. It needs to work properly.

Additionally, Helm methods don't support context, so we are not able to cancel any ongoing operation even if user cancelled the request and no one will wait for the response.

## Related issue(s)

Fix https://github.com/capactio/capact/issues/665


deploy/kubernetes/charts/helm-storage-backend/values.yaml
```diff
@@ -15,12 +15,30 @@ helmReleaseBackend:
...
 helmTemplateBackend:
   enabled: true
   service:
     port: 50052
     type: ClusterIP
 
+  resources:
+    limits:
+      cpu: 1
+      memory: 512Mi
```
pkosiec, line 36: tenor-218226211

internal/helm-storage-backend/template.go
```diff
@@ -2,32 +2,214 @@ package helmstoragebackend
...
+	// It is important to load the chart dependencies as by default they are not
+	// available and if chart was using that we may miss some value and template funcs.
+	chartDeps, err := h.loadHelmChartDependencies(helmRelease.Chart)
+	if err != nil {
+		return nil, err
+	}
+
+	helmRelease.Chart.SetDependencies(chartDeps...)
+	if err := chartutil.ProcessDependencies(helmRelease.Chart, helmRelease.Config); err != nil {
+		return nil, err
```
pkosiec, line 173: I think we should wrap the error with `gRPCInternalError` in these places as well

internal/helm-storage-backend/template_test.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+}
+
+func TestTemplate_CreateGetAndUpdate_Failures(t *testing.T) {
+	// globally given
+	const (
+		releaseName      = "test-release"
+		releaseNamespace = "test-namespace"
+	)
+	tests := []struct {
```
pkosiec, line 135: It would be great to test also output when user provided a wrong template

pkosiec (approved):
Works well, I've just tested that E2E on my PR (https://github.com/capactio/hub-manifests/pull/64). Just a few minor comments 🙂

PR #696, opened 2022-04-05
Investigate nested workflows dynamic ti

### LINKED ISSUE #688: Investigate dynamic TypeInstances support in umbrella workflows
## Description

Currently, our PostgreSQL installation Implementation produces static TypeInstance for PostgreSQL Helm release ([link](https://github.com/capactio/hub-manifests/blob/b44676af4956408b49c02901b5c52d6a1466bd77/manifests/implementation/bitnami/postgresql/install.yaml#L118)). The reason is the current limitation of accessing our dynamic TypeInstances in umbrella workflows.

In this case, the umbrella workflow is the Mattermost installation ([link](https://github.com/capactio/hub-manifests/blob/b44676af4956408b49c02901b5c52d6a1466bd77/manifests/implementation/mattermost/mattermost-team-edition/install.yaml#L121)). 

If we used the Helm template storage, the `postgresql` artifact would contain:

```
backend:
  context:
    goTemplate: |
      host: '{{ template "postgresql.primary.fullname" . }}'
      port: '{{ template "postgresql.port" . }}'
      defaultDBName: '{{ template "postgresql.database" . }}'
      superuser:
        username: '{{ template "postgresql.username" . }}'
        password: '{{ template "postgresql.password" . }}'
    release:
      driver: secrets
      name: example-release
      namespace: default
value: null
``` 

And the Mattermost installation umbrella workflow steps require these rendered values, to:
- [create DB user](https://github.com/capactio/hub-manifests/blob/b44676af4956408b49c02901b5c52d6a1466bd77/manifests/implementation/mattermost/mattermost-team-edition/install.yaml#L130)
- [create DB](https://github.com/capactio/hub-manifests/blob/b44676af4956408b49c02901b5c52d6a1466bd77/manifests/implementation/mattermost/mattermost-team-edition/install.yaml#L159)

## AC

- Summarize options that we have to solve this issue
- Brainstorm and agree with team what we want to check as a part of PoC
- If possible, create a PoC of a given solution

### Ideas how to solve it:

1. Change the way how we run nested workflow: maybe the Engine shouldn't render one huge workflow, but schedule and run separate Argo Workflows instead? Every workflow would have dedicated upload/download steps
    - Investigated as a part of #546
3. Save rendered value under different property in the artifact (e.g. `.rendered`) and use that in umbrella workflow
    - Ignore it during TypeInstance upload
    - It probably won't work for other backends, which base on actual TypeInstance existing in Hub
1. Hack: Add "fetcher" container which fetches Helm template storage backend and renders the template
    - We would need to fake TypeInstance ID, because there's no TypeInstance yet
    - It won't work for other backends, which base on actual TypeInstance existing in Hub
    - Probably no-go

## Reason and use cases

See https://github.com/capactio/capact/issues/604

pkosiec:
## Description

- Describe approach for dynamic TypeInstances from nested workflows
- Update gRPC API for Storage Backends and its usage

## Related issue(s)

Resolves #688


docs/proposal/20211207-delegated-storage.md
```diff
@@ -536,42 +574,59 @@ Capact Local Hub calls proper storage backend service while accessing the TypeIn
...
+      rpc OnUpdate(OnUpdateValueAndContextRequest) returns (OnUpdateResponse);
+      rpc OnDelete(OnDeleteValueAndContextRequest) returns (OnDeleteResponse);
+    
+      // lock
+      rpc GetLockedBy(GetLockedByRequest) returns (GetLockedByResponse);
+      rpc OnLock(OnLockRequest) returns (OnLockResponse);
+      rpc OnUnlock(OnUnlockRequest) returns (OnUnlockResponse);
+    }
+    
+    // ContextOnlyStorageBackend handles TypeInstance lifecycle based on the context, which is required. TypeInstance value is never passed in input arguments.
```
mszostok, line 608: IMO there is no need to say `Only` the:
- `ValueAndContextStorageBackend`
- and `ContextStorageBackend`
names, are more consistent.

mszostok (approved):
LGTM. In general in didn't dive into the topic about service composition but it's quite strange that there is no such concept as I see a lot of cases where such thing can be used. 

For now, it fulfills our needs, so I would not spend more time on it. Later, if we will have a real use case we can revisit it.

Issue #661, opened 2022-03-13
Always store the `value` property on `update` mutation


mszostok:
## Description

Currently, we don't update the `value` property on `update` mutation when the `value` field is not requested to be returned on response. This is not acceptable, and we need to fix it. We were not able to solve this problem in https://github.com/capactio/capact/issues/645 as we run of time.

## Options

We pick option 1:

**1. Move code to the `update` mutation resolver, but this will make the transaction handling harder. This will be done before the query will be executed. In case of failed query, we won't be able to remove it from external storage.**
- Add DeleteRevision for storage backends
- Update docs to mention new gRPC method

~1. On `update` mutation, return only `id` and `alias`. Same as we do on `create` mutation. It is  the simplest solution, but significantly degrades usability.~
    - in Go client, run update + get in a single HTTP call

~1. Manipulate the query fields. If user didn't ask for `value` add it on request, and remove it on response.~
 - This is the most hack-ish way of solving this problem.

## Expected behavior

The `value` property is always updated, and it doesn't depend on the user query return fields.

## Actual behavior

If the `value` is not requested on the GQL response, the `value` is not updated.

## Related issues

See epic #604 for reason and use cases.

PR #698, opened 2022-04-06
Add opt to remove TypeInstance's revision, fix update mutation

mszostok:
## Description

- Add opt to remove TypeInstance's revision via dedicated gRPC method
- Store new revision in external storage before executing update cypher query
- Rollback stored revision in case of failed transaction

The Local Hub code was simplified thanks to handling the update logic directly in mutation resolver instead of `value` field resolver. However, the additionall logic (**rpc DeleteRevision**) is now required to be implemented on each storage backend.

## Testing

I added e2e tests for Local Hub to prove that the contract was not broken with the new approach.

## Related issue(s)

Fix https://github.com/capactio/capact/issues/661


internal/secret-storage-backend/server_test.go
```diff
@@ -656,6 +657,146 @@ func TestHandler_OnDelete(t *testing.T) {
...
+				"cant-touch-this": {
+					"Music":        "hits me so hard",
+					"Makes me say": "Oh, my Lord",
+				},
+			}),
+			ExpectedProviderState: map[string]map[string]string{
+				path: {
+					"1": "key-one-value",
+				},
+				"cant-touch-this": {
```
mkuziemko, line 713: :+1:

internal/secret-storage-backend/server.go
```diff
@@ -268,6 +269,40 @@ func (h *Handler) OnDelete(_ context.Context, request *pb.OnDeleteRequest) (*pb.
...
+// OnDeleteRevision handles TypeInstance's revision deletion by removing a secret entry in a given provider.
+// It checks whether a given TypeInstance is locked before doing such operation.
+func (h *Handler) OnDeleteRevision(ctx context.Context, request *pb.OnDeleteRevisionRequest) (*pb.OnDeleteRevisionResponse, error) {
+	if request == nil {
+		return nil, NilRequestInputError
+	}
+
+	provider, err := h.getProviderFromContext(request.Context)
+	if err != nil {
+		return nil, err
```
mkuziemko, line 281: should we use `errors.Wrap`?

mkuziemko (commented):
I verified it locally and it works well :+1: small comments only

test/local-hub/external_storage_backend_test.go
```diff
@@ -412,6 +415,120 @@ func TestExternalStorage(t *testing.T) {
...
+// TestDeleteTypeInstanceRevision can be run in parallel with other tests but
+// the subtests cannot be executed parallel.
+//
+// GRPC_SECRET_STORAGE_BACKEND_ADDR=50051 go test --tags=localhub -run=TestDeleteTypeInstanceRevision ./test/local-hub/...
+func TestDeleteTypeInstanceRevision(t *testing.T) {
+	// globally given
+	srvAddr := os.Getenv(storageBackendAddr)
+	if srvAddr == "" {
+		t.Skipf("skipping running test as the env %s is not provided", storageBackendAddr)
+	}
```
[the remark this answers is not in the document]
mszostok: IMO it won't be worth it. As there are two options:
- don't pass any input args, but func needs to return multiple of them even though it's kinda not its responsibility
- extract and pass external args, this won't simplify the code

internal/secret-storage-backend/server.go
mszostok: nope, in other places we also don't use it as the `getProviderFromContext` already returns proper grpc errors, and we don't want to override them

mkuziemko (approved):
LGTM :+1:

Issue #546, opened 2021-10-29
Composed Action spike


lukaszo:
## Description

Sometimes we need to create so-called "Umbrella" manifests where we want to deploy several applications. This requires creating one Interface, InterfaceGroup, Types and so on. The unique part about it is that it has only one Implementation. 
The only purpose of this implementation is to group actions in one place and pass artifacts between them.
This requires a lot of boilerplate YAML. 

The solution may be to allow to run Composed Actions(CA) directly from API, without requiring them to exist in hub.
The syntax of CA would be very similar to the Implementation. It could have embedded policy and input. 
It could be possible to store CA in files on remote locations and run actions by just providing a link to the file. 

This would be similar to what Docker Compose is to Docker.

Problems:
* Composed actions may become big. This will require creating some guideline which tells when to use what
* It may also require defining temporal Types

Alternatives:
* Allow creating Implementations without Interfaces

To discuss:
- Do we need Interfaces for such Implementations? Probably not
- Should they be stored in Hub? Probably yes

## Reason

* Improve  User Experience
* Reduce number of "Umbrella" manifests in hub

## Use cases

System user will be able to create advanced deploy scenarios without having access to the HUB.

mszostok:
Could solve also: https://github.com/capactio/capact/issues/563

PR #701, opened 2022-04-11
Add composed actions raw notes from a short investigation

mszostok:
## Description

- Add composed actions raw notes from a short investigation

## Related issue(s)

- https://github.com/capactio/capact/issues/546


docs/investigation/composed-action/REAMDE.md
```diff
@@ -0,0 +1,163 @@
...
+        path: "cap.interface.database.postgresql.install"
+      input:
+        input-parameters:
+          raw:
+            data: |
+              superuser:
+                username: superuser
+              defaultDBName: postgres
+    create-user:
+      depends: ["psql"] # this can be removed, we can create own DAG based on the inputs.
```
pkosiec, line 56: Maybe we can rename this feld to `dependencies`, `depends-on` or `dependsOn` or `needs` (like GH Actions). It will be more readable 🙂 

```suggestion
      depends: ["psql"] # this can be removed, we can create own DAG based on the inputs.
```

docs/investigation/composed-action/REAMDE.md
```diff
@@ -0,0 +1,163 @@
...
+        user:
+          from: create-user.user
+  advancedRendering:
+    enabled: false
+  run: false
+  dryRun: false
+  cancel: false
+```
+
+In the future, we can also add a native support for direct Implementation references in steps. When user compose an Action, they already know what they want to achieve. Instead of creating a complex Action Policy, specific Implementations can be selected with an implementation-specific input attached to them. In this way, the Interfaces would be a nice grouping feature used on UI and later user by browsing a catalog of available service, can compose a given Action. Such Action will be easier to rationalize and requires less cognitive overhead as it describes in a declarative way want will be done.
```
pkosiec, line 97:
```suggestion
In the future, we can also add a native support for direct Implementation references in steps. When users compose an Action, they already know what they want to achieve. Instead of creating a complex Action Policy, specific Implementations can be selected with an implementation-specific input attached to them. In this way, the Interfaces would be a nice grouping feature used on UI and later user by browsing a catalog of available service, can compose a given Action. Such Action will be easier to rationalize and requires less cognitive overhead as it describes in a declarative way want will be done.
```

docs/investigation/composed-action/REAMDE.md
```diff
@@ -0,0 +1,163 @@
...
+
+**Follow-ups**
+- Add support for parallel and sequence execution,
+- Add support for error handling and retries,
+- Add support to store templates in Public Hub,
+- Add support for direct Implementation references in steps,
+- Add support to create ComposedAction via Dashboard:
+  - Later a dedicated block builder can be added.
+
+### Summary
```
pkosiec, line 161: I think it will be more convenient if you also copy-paste the Decision from the PR description into this doc 🙂

pkosiec (approved):
LGTM - please put the decision in this doc and we can merge it 🙂

docs/investigation/composed-action/REAMDE.md
mszostok: The `depends` was taken from the `argo` syntax. Personally, I don't have a strong opinion about the property name. I added all alternatives to the doc 👍

Issue #504, opened 2021-09-17
Change output.goTemplate in Terraform-based Implementations to be of type string


mszostok:
## Description

We had to change the `output.goTemplate` in Helm runner to be a string and not a map. 
To keep it consistent we have to change it also for the Terraform runner and Terraform-based Implementations e.g. `cap.implementation.gcp.cloudsql.postgresql.install`.

## AC
- Change Terraform runner `output.goTemplate` property to string: https://github.com/capactio/capact/blob/main/pkg/runner/terraform/types.go#L36

- Ensure render logic works properly: https://github.com/capactio/capact/blob/main/pkg/runner/terraform/terraform.go#L163-L169
- Modify the Terraform-based Implementations. It should be sufficient to make the following changes:
    ```yaml
    output:
      goTemplate:
        host: "..."
    ```
    to
    ```yaml
    output:
      goTemplate: |
        host: "..."
    ```

## Reason

["Consistent code is easier to maintain, is easier to rationalize, requires less cognitive overhead, and is easier to migrate or update as new conventions emerge or classes of bugs are fixed."](https://github.com/uber-go/guide/blob/master/style.md#be-consistent)

mszostok:
**Issue overview:**

<table>
<tr>
<td> Before </td> <td> After </td>
</tr>
<tr>
<td>

  ```yaml
  output:
    goTemplate:
      host: "{{ .instance_ip_addr }}"
      port: {{ .port }}
      defaultDBName: "{{ .defaultDBName }}"
      superuser:
        username: "{{ .username }}"
        password: "{{ .password }}"
      identifier: "{{ .identifier }}"
      availability_zone: "{{ .availability_zone }}"
      class: "{{ .class }}"
  ```

</td>
<td>

  ```yaml
  output:
    goTemplate: |
      host: "{{ .instance_ip_addr }}"
      port: {{ .port }}
      defaultDBName: "{{ .defaultDBName }}"
      superuser:
        username: "{{ .username }}"
        password: "{{ .password }}"
      identifier: "{{ .identifier }}"
      availability_zone: "{{ .availability_zone }}"
      class: "{{ .class }}"
  ```

</td>
</tr>
</table>

This resolves the rendering issue:

```
while unmarshaling data from file "/terraform-args": error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".port":interface {}(nil)}
```

Issue #666, opened 2022-03-15
Adjust all Implementation manifests to use latest workflow syntax and storage backends


pkosiec:
## Description

Blocked by #634, #650, #665

- Adjust all Implementation manifests inside `capactio/hub-manifests` to use the latest workflow syntax (Argo workflows artifact content) - breaking change introduced in #634 
- Use Helm storage for all Helm installation (introduced in #650)
- Use dynamic TypeInstance projection for output TypeInstances (#665 and #650)
- Update RocketChat and Mattermost installation tutorial (add prerequisite for Helm storage)

**Additional scope:**
- Move outputting static Helm Release logic from Helm Runner inputs to env variables
- Remove `output.useHelmReleaseStorage` from `helm.install` and add env variable to Helm Runner container. We have now two Implementations (`install` and `install-static`, so we can distinguish them). However, all manifests that use Helm runner should be updated to always assume they use the new `helm.install` which outputs dynamic Helm Release TI. That means:

```yaml
capact-outputTypeInstances:
  - name: helm-release
     from: helm-release
```

Needs to be changed to:

```yaml
capact-outputTypeInstances:
  - name: helm-release
     from: helm-release
     backend: helm-release-storage # from requires section
```

- Remove `output.useHelmReleaseStorage` from `helm.upgrade` input Type - detect input TypeInstance in Helm Runner and output the same shape
- Consider making a dedicated Interface and Implementation for the `ti-value-fetcher`.

Timebox for 3SP.

## Related issues

See epic https://github.com/capactio/capact/issues/604 for reason and use cases.

mszostok:
Testing scenarios:

Helm storage + AWS Storage

AWS:
- Elasticsearch - tested separately ✅ (spot issue with version and host name)
- Redis - tested separately
- secrets-manager - tested by ??
- RDS - tested as a part of Mattermost install

Bitnami:
- MinIO - tested separately
- PostgreSQL - tested separately
- Redis - tested separately ✅ (spot issue with high memory consumption on Helm Template storage (>1.5G))
- MongoDB - tested as a part of RocketChat install ✅

Capact:
-  Upgrade - the Helm storage is not available when the Capact is installed, as a result, we cannot change Upgrade action to dynamic Helm storage.

Elastic:
- Elasticsearch - tested separately ✅

GCP:
- CloudSQL - tested as a part of Concourse install


1. Install Mattermost
   1. on-prem (Helm chart) PostgreSQL
   2. RDS
[...]

PR #707, opened 2022-04-21
Narrow-down the download TypeInstance details

mszostok:
## Description

- Narrow-down the download TypeInstance details. Currently, we download all details about TypeInstance together with all relations. This sometimes causes timeouts in download step. Important is that we don't need those details. 
- Change goTemplate to string in Terraform runner
- Add required delay for value fetcher

## Related issue(s)

- https://github.com/capactio/capact/issues/666
- Fix https://github.com/capactio/capact/issues/504


mkuziemko (approved):
LGTM :+1:

mkuziemko (approved):
Still, LGTM, although the title of the PR could be more accurate :)

PR #718, opened 2022-05-11
Add pipeline to run integration test on k3d

### LINKED ISSUE #492: Add pipeline to run integration test on k3d
## Description

Recently, we added support for k3d (https://github.com/capactio/capact/issues/466). To ensure no regression, we need to create a pipeline that will run on each PR and main branch. It will be exactly the same as we have with the kind.

https://github.com/capactio/capact/blob/fa6712d8d3fc5767e063a763dc7717eb9ca22b3c/.github/workflows/pr-build.yaml#L390-L443

We need to add it as a parallel step to the `kind` one.

## AC

- Integration test are executed on k3d cluster, if failed, pipeline should be red
- Integration test on kind and k3d are run in parallel

- **Stretch:** 
    Use preloaded images for our dependencies. We should build image tarball only if something was changed in `deploy/kubernetes/charts`. Later, this tarball can be used by k3d. 
    ```
    # Cache images
    capact alpha archive-images helm -v --version @latest --helm-repo @latest --output ./archive.tar --save-component=neo4j,ingress-nginx,argo,cert-manager,kubed
    
    # Preload images
    DISABLE_MONITORING_INSTALLATION=true K3D_IMAGES_ARCHIVE=archive.tar CLUSTER_TYPE=k3d make test-integration
    ```
    
    Not sure where this tarball should be stored, as we may run to GitHub cache limits: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy. Maybe it will be better to push it to our GCS.

## Reason

Ensure no regression. Additionally, test which environment is more stable and faster.

blazejmalecki:
## Description

- Enable k3d cluster type in integration tests pipeline

Pipeline with enabled k3d integration tests passes - https://github.com/blazejmalecki/capact/runs/6383579097?check_suite_focus=true.

## Related issue(s)

- Resolves #492

### DIFF
--- a/.github/workflows/branch-build.yaml
+++ b/.github/workflows/branch-build.yaml
@@ -281,11 +281,12 @@ jobs:
           chmod +x ./capact
       - name: Run cross-functional integration tests
         env:
-          BUILD_IMAGES: "false"
+          BUILD_IMAGES: "true"
           ARTIFACTS: "output/"
           DISABLE_MONITORING_INSTALLATION: "true"
           CAPACT_BINARY: "./capact"
           CLUSTER_TYPE: ${{ matrix.cluster }}
+          K3D_REGISTRY: "--enable-registry"
         run: |
           make test-integration
       - name: Upload artifacts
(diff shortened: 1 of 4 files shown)


mszostok (approved):
LGTM 👍

Issue #615, opened 2022-01-25
Add option to get logs from executed Actions


mszostok:
## Description

Currently, we are not able to get logs from executed Actions via Capact CLI. As a result, user needs to use the `kubectl` and get a proper pods names.

We can distinguish between two options:
- get the latest logs from the Action execution
- stream logs from the Action execution

The first one can be already implemented, where for the second one it will be good to use the websockets. Unfortunately, we don't support it yet. 

## Reason

Requires installation of external tools (kubectl and configure it with proper cluster) for some of our examples. This should be simplified and Capact CLI should be the only one which is needed. One 💍  to rule them all 😎 . Same as we did for #328 and #466.

## Use cases

- **System User** - with Capact CLI can easily get the Action logs. Currently, it involves usage of external CLIs and manual discovery of related pods.

mszostok:
The first implementation was added. However, it's done via direct call to k8s cluster. This should be abstracted and calls should be done via Gateway (GraphQL or REST endpoint). Because of that, I'm not closing the issue yet.

PR #722, opened 2022-05-13
Add act logs command

mszostok:
## Description

- Add act logs command
- I had to change the short form for the `--config` flag. Now it's `-C`. IMO it's better as in k8s world on which we rely for now the `-c` is popular short form for the `--container` flag.

## Testing

1. Checkout PR
2. Build CLI `make build-tool-cli`
3. Login into long-running cluster
4. Create `cap.interface.capactio.examples.greet:0.1.0` Action.
    ```bash
    capact act create cap.interface.capactio.examples.greet --name example
    capact act wait --for=phase=READY_TO_RUN example
    ```
6. Run the Action:
    ```bash
    capact act run example
    ```
7. Stream logs:
    ```bash
    capact act logs example -n default -f
    ```

After it's finished, you can play with other options:
```bash
capact act logs @latest -n default
capact act logs example -n default --timestamps
capact act logs example -n default --grep "Hello"
capact act logs example -n default --since 10m
capact act logs example -n default --since 2s
capact act logs example -n default --no-color
```
## Related issue(s)

- https://github.com/capactio/capact/issues/615 - IMO the issue should not be closed until we change the implementation to become k8s agnostic (via Gateway with GraphQL or REST)

### DIFF
--- a/cmd/cli/cmd/action/logs.go
+++ b/cmd/cli/cmd/action/logs.go
@@ -1,0 +1,50 @@
+package action
+
+import (
+	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands"
+	"github.com/argoproj/argo-workflows/v3/cmd/argo/commands/client"
+
+	"capact.io/capact/internal/cli"
+	"capact.io/capact/internal/cli/heredoc"
+
+	"github.com/spf13/cobra"
+)
+
+// NewLogs returns a new cobra.Command for getting Action's logs.
+// TODO: this should be done via Gateway once the subscription is implemented.
+func NewLogs() *cobra.Command {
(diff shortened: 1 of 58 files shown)


cmd/cli/cmd/action/logs.go
```diff
@@ -0,0 +1,50 @@
+package action
```
pkosiec, line 1: Please synchronize the docs to website, and make sure the change from `-c` to `-C` is applied onto our docs (assuming we used this flag somewhere in the documents). Also, replace `argo logs` usage with `capact act logs` in all docs. Thanks!

pkosiec (approved):
👌

PR #726, opened 2022-05-19
Add support for overrides to CLI upgrade

### LINKED ISSUE #497: Support custom overrides for Capact CLI upgrade
## Description

Currently, you can specify only a few overrides for Capact upgrade via dedicated flags, e.g. `override-capact-image-tag`.

The goal of this task is to allow custom overrides for all components. This should be done in the same way as we did for `capact install` to have consistent approach:

https://github.com/capactio/capact/blob/de7fae1fd83324669798daa3f1c7e1c76b2ea66d/cmd/cli/cmd/install.go#L66-L69

As a result, it can be used in form `capact upgrade --capact-overrides="hub-public.populator.manifestsLocation.repository=Foo,hub-public.populator.manifestsLocation.branch=Bar`
## AC

- Handle custom overrides per component in the same fashion as we do for `capact install`
- Merge values with our overrides (e.g. for increased resource limits)

## Reason

We need to have a way to provide overrides via CLI to e.g. set Populator branch to new release branch from `release-0.2` to `release-0.3` during upgrade.
Here we were not able to specify such command: https://github.com/capactio/capact/releases/tag/v0.3.0.

blazejmalecki:
## Description

- Add support for overrides to CLI `upgrade` command

## Related issue(s)

- Resolves #497


mkuziemko:
@blazejmalecki could you describe our ideas/findings in the GitHub issue? It would be good to keep this on track. Thanks! :+1:

